PR #68, opened 2024-12-12
feat: EMLX compiled mode

### LINKED ISSUE #61: Use mlx::compile for the Nx.Defn compiler
MLX uses an abstraction for compiling Python functions: https://github.com/ml-explore/mlx/blob/974bb54ab2f8e8450f856ac215df160374080b33/python/src/transforms.cpp#L370C8-L370C21

We can copy most of that workflow to get a similarly-compiled defn function in EMLX.__compile__. EMLX can get the (vector<array>) -> vector<array> function pointer as a NIF resource, cache that in the compiler key and use it for the Elixir function as well.

polvalente:
This PR uses Cocoa's `nif_call` library to bridge the missing
gap and implement a proper Nx.Defn compiler for EMLX.

closes #61

Benchmark of backend [EMLX, EXLA] against compiler [self, Nx.Defn.Evaluator]
```elixir
Mix.install [{:emlx, path: __DIR__}, :benchee, :exla]

defmodule ToBench do
  def run(backend, compiler) do
    prev_backend = Nx.default_backend(backend)
    fun = fn x, y ->
      Enum.reduce(1..10, x, fn x, acc ->
        Nx.dot(x, acc)
        |> Nx.add(y)
      end)
    end


    Nx.Defn.jit_apply(fun, [Nx.iota({10, 10}, type: :f32, backend: backend), Nx.iota({10, 10}, type: :f32, backend: backend)], compiler: compiler)
    |> tap(fn _ -> Nx.default_backend(prev_backend) end)
  end
end

Benchee.run(%{
  "EMLX (evaluator)" => fn -> ToBench.run(EMLX.Backend, Nx.Defn.Evaluator) end,
  "EMLX" => fn -> ToBench.run(EMLX.Backend, EMLX) end,
  "EMLX gpu" => fn -> ToBench.run({EMLX.Backend, device: :gpu}, EMLX) end,
  "EXLA (evaluator)" => fn -> ToBench.run(EXLA.Backend, Nx.Defn.Evaluator) end,
  "EXLA" => fn -> ToBench.run(EXLA.Backend, EXLA) end
})
```

```
Name                       ips        average  deviation         median         99th %
EXLA                   37.97 K       26.34 μs    ±49.15%          22 μs       94.71 μs
EMLX                   16.64 K       60.08 μs    ±43.45%       52.88 μs      204.13 μs
EMLX gpu               16.11 K       62.07 μs    ±61.75%       53.29 μs      206.79 μs
EMLX (evaluator)        3.48 K      287.49 μs    ±30.55%      280.96 μs      421.05 μs
EXLA (evaluator)        3.17 K      315.41 μs    ±28.33%      296.25 μs      540.18 μs

Comparison: 
EXLA                   37.97 K
EMLX                   16.64 K - 2.28x slower +33.75 μs
EMLX gpu               16.11 K - 2.36x slower +35.73 μs
EMLX (evaluator)        3.48 K - 10.92x slower +261.16 μs
EXLA (evaluator)        3.17 K - 11.98x slower +289.08 μs
```


c_src/emlx_nif.cpp
```diff
@@ -479,6 +507,73 @@ NIF(eval) {
...
+    if (!nx::nif::get_list(env, output_list, output_tensors)) {
+      throw std::runtime_error("Failed to convert callback result to tensors");
+    }
+
+    enif_free_env(env);
+
+    return output_tensors;
+  };
+
+  emlx::function compiled_function_ptr = mlx::core::compile(fun);
```
polvalente, line 559: @awni I got things working via the idea we discussed yesterday! ~5x speed up vs the non-compiled version as per the benchmark in the PR description.

I'm do wonder about what XLA is doing to get such a performance boost when compiled.

awni: Cool, so you figured out how to pass an elixir function into C/C++?

cocoa-xu (approved):
Pushed 2 commits and update nif_call, it should work now

polvalente:
Merging this! We still need to allow BinaryBackend tensors to be automatically allocated if passed as inputs.
Otherwise, we're ready for v0.1.0

README.md
```diff
@@ -40,7 +40,10 @@ Nx.default_backend({EMLX.Backend, device: :cpu})
 Nx.default_backend({EMLX.Backend, device: :gpu})
 ```
 
-If desireable, you can also set the default compiler:
+If you want to use the JIT compiler, you can set the default compiler as shown below.
+
+Currently, the underlying implementation is not totally safe and could lead to a deadlocked dirty NIF, so this must be used with care.
```
polvalente, line 45: @cocoa-xu is this still valid?

cocoa-xu: I think we can remove this. It's actually not because of a deadlock, it was the registered tag was unregistered too early. And speaking of this, there needs to be a follow-up PR to manually unregister the tag in `nif_call`. I'll create an issue for this.

PR #74, opened 2025-02-13
feat: added experimental support for linux

cocoa-xu:


cocoa-xu:
After some debugging, this appears to be an issue in `EMLX.astype` for x86_64 linux when using CPU backend.

```elixir
[lib/emlx/backend.ex:1224: EMLX.Backend.as_type/2]
EMLX.to_blob(t) #=> <<0, 0, 128, 127>>

[lib/emlx/backend.ex:1225: EMLX.Backend.as_type/2]
to_nx(EMLX.astype(t, to_mlx_type(type)), out) #=> #Nx.Tensor<
  u8
  0
>

[lib/emlx/backend.ex:1226: EMLX.Backend.as_type/2]
EMLX.to_blob(EMLX.astype(t, to_mlx_type(type))) #=> <<0>>
```

Going to skip that unit test.

PR #75, opened 2025-02-14
chore(ci): test gpu backend on ci

cocoa-xu:


cocoa-xu:
Getting some errors in `test/emlx/nx_test.exs`:

```
----test/emlx/nx_test.exs----

LIBMLX_ENABLE_DEBUG=false
Terminating due to uncaught exception: [metal::Device] Unable to load kernel gg0_copyint32int32

bash: line 1:  1914 Abort trap: 6           mix test $NAME
```

Might be an upstream issue because the same error also presents when building from source on CI.

PR #79, opened 2025-08-13
remove raising on tie_break :high

### LINKED ISSUE #78: Warn instead of raising in argmax for tie_break: :high
Hello, great project!

I'm trying to run `facebook/bart-large-mnli` with `EMLX` which results in 
```bash
** (RuntimeError) Nx.Backend.argmax/3 with tie_break: :high is not supported in EMLX
```

for quick reproduction:

```elixir
Mix.install([{:emlx, github: "elixir-nx/emlx"}, {:bumblebee, "~> 0.6.0"}])

Nx.global_default_backend(EMLX.Backend)

{:ok, model} = Bumblebee.load_model({:hf, "facebook/bart-large-mnli"})
{:ok, tokenizer} = Bumblebee.load_tokenizer({:hf, "facebook/bart-large-mnli"})

labels = ["happy", "sad"]

serving = Bumblebee.Text.zero_shot_classification(model, tokenizer, labels)

text = "hello, I am happy."

result = Nx.Serving.run(serving, text)

IO.puts(inspect(result))
```

relevant code in `EMLX.Backend`:

```elixir
  # Aggregation (axis)
  ops = [:argmax, :argmin]

  for op <- ops do
    @impl true
    def unquote(op)(out, tensor, opts) do
      axis = opts[:axis]
      keep_axis = opts[:keep_axis] == true

      if opts[:tie_break] == :high do
        raise "Nx.Backend.#{unquote(op)}/3 with tie_break: :high is not supported in EMLX"
      end

      t_mx = from_nx(tensor)

      result =
        if axis do
          EMLX.unquote(op)(t_mx, axis, keep_axis)
        else
          EMLX.unquote(op)(t_mx, keep_axis)
        end

      result
      |> EMLX.astype(to_mlx_type(out.type))
      |> to_nx(out)
    end
  end
```

I'm assuming that `EMLX` will always use low tie break given that high is not supported.

So, as I'm not directly controlling the model code if I'm not forking `bumblebee`, could we just emit a warning that `tie_break: :high` is not supported and instead it will be `:low`, which might lead to unexpected results but let users still run the model?

joelpaulkoch:
Removes raising on tie_break: :high option in argmin and argmax to allow to run models that use this option, closes #78.

### DIFF
--- a/lib/emlx/backend.ex
+++ b/lib/emlx/backend.ex
@@ -535,10 +535,6 @@ defmodule EMLX.Backend do
       axis = opts[:axis]
       keep_axis = opts[:keep_axis] == true
 
-      if opts[:tie_break] == :high do
-        raise "Nx.Backend.#{unquote(op)}/3 with tie_break: :high is not supported in EMLX"
-      end
-
       t_mx = from_nx(tensor)
 
       result =


joelpaulkoch:
I just removed the raise. Not sure if we could warn in a reasonable way, simply using `Logger` would warn everytime the function runs, right? So I guess that's too noisy.

polvalente:
Perhaps we could add that behind an EMLX-specific :warn_unsupported_option config that defaults to true.

PR #95, opened 2026-01-22
feat: Add quantization operations (quantized_matmul, dequantize, quantize)

notactuallytreyanastasio:
## Summary

Add MLX quantization NIFs for efficient 4-bit model inference:

- `quantized_matmul`: Multiplies input with quantized weights (uint32 packed int4)
- `dequantize`: Converts quantized weights back to floating point  
- `quantize`: Quantizes float tensor to packed format (returns `{weights, scales, biases}`)

## Motivation

These operations enable running 4-bit quantized LLMs like Qwen3-8B-4bit, LLaMA-4bit, etc. with the EMLX backend. The `quantized_matmul` operation is the key primitive for efficient inference with memory-constrained models on Apple Silicon.

## API

```elixir
# Quantize weights for storage/inference
{q_weights, scales, biases} = EMLX.quantize(weights, 64, 4)

# Efficient inference with quantized weights
output = EMLX.quantized_matmul(input, q_weights, scales, biases, true, 64, 4)

# Dequantize for debugging/verification
recovered = EMLX.dequantize(q_weights, scales, biases, 64, 4)
```

## Performance

Tested with 4-bit Qwen3-8B model inference:
- 135 tok/s single-token inference
- ~21 tok/s text generation

On Apple Silicon M-series chips.

## Test plan

- [x] Added `test/emlx/quantization_test.exs` with 6 tests covering:
  - Basic quantization of float tensors
  - Shape verification for quantized outputs
  - Dequantization back to float
  - Quantize-dequantize roundtrip (approximate equality)
  - Quantized matmul shape verification
  - Quantized matmul produces reasonable values

- [x] All existing 2117 tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/c_src/emlx_nif.cpp
+++ b/c_src/emlx_nif.cpp
@@ -970,6 +970,63 @@ NIF(as_strided) {
   TENSOR(mlx::core::as_strided(*t, shape, strides, offset, device));
 }
 
+// ============================================================================
+// Quantization Operations (for 4-bit model support)
+// ============================================================================
+
+// quantized_matmul - Multiplies x with a quantized weight matrix w
+// This is the key operation for efficient 4-bit inference
+// MLX API: quantized_matmul(x, w, scales, biases, transpose, group_size, bits, stream)
+NIF(quantized_matmul) {
+  TENSOR_PARAM(0, x);       // Input tensor [batch, seq, hidden]
+  TENSOR_PARAM(1, w);       // Quantized weights [out/8, in] (uint32 packed)
+  TENSOR_PARAM(2, scales);  // Scales [out/group_size, in] (bfloat16)
+  TENSOR_PARAM(3, biases);  // Biases [out/group_size, in] (bfloat16)
+  PARAM(4, bool, transpose);
+  PARAM(5, int, group_size);
+  PARAM(6, int, bits);
+  DEVICE_PARAM(7, device);
+
+  TENSOR(mlx::core::quantized_matmul(
+      *x, *w, *scales, *biases, transpose, group_size, bits, device));
+}
+
+// dequantize - Converts quantized weights back to float
+// Useful for debugging and verification
+// MLX API: dequantize(w, scales, biases, group_size, bits, stream)
+NIF(dequantize) {
+  TENSOR_PARAM(0, w);       // Quantized weights (uint32 packed)
+  TENSOR_PARAM(1, scales);  // Scales (bfloat16)
+  TENSOR_PARAM(2, biases);  // Biases (bfloat16)
+  PARAM(3, int, group_size);
+  PARAM(4, int, bits);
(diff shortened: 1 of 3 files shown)


polvalente:
@notactuallytreyanastasio is there a reason why you closed your PR? These might be welcome additions, but we need to understand how these would actually be useful.

For instance, we might need some type of config to override Nx.dot when type is s4 instead of exposing a separate EMLX-only operation

notactuallytreyanastasio:
@polvalente I opened it a bit prematurely and wanted to write things up better, but Claude mistakenly updated it and sent it here. I will open a full write up and give my use case in the next couple days.

polvalente:
> @polvalente I opened it a bit prematurely and wanted to write things up better, but Claude mistakenly updated it and sent it here. I will open a full write up and give my use case in the next couple days.

Thank you! Looking forward to it. If you do wanna explore something along the lines of what I think is gonna be needed, you can try adding a backend option for tensors in EMLX that will annotate whether the tensor is, for instance, s4 quantized.
And then operations such as Nx.dot can branch based on that do invoke the proper quantized matmul implementations.

There might also be something to configure at the MLX level to do this branching autonatically.

notactuallytreyanastasio:
@polvalente I have some work almost complete for all of that, see these links, I also threw in some LoRA stuff that does a few clever tricks

https://github.com/notactuallytreyanastasio/bumblebee_quantized

https://github.com/notactuallytreyanastasio/safetensors_ex

https://github.com/notactuallytreyanastasio/emlx/tree/feat/quantization-ops

https://github.com/notactuallytreyanastasio/bobby_posts_adapters

notactuallytreyanastasio:
I was basically just trying to, in pure elixir, replicate a setup with qwen3 that I trained on 250,000 of my twitter posts and 25,000 of my Bluesky posts and make it post like me, and this was what it took cuz it was a 4 bit model

polvalente:
I think if the only thing missing if a way to replace Nx.dot with EMLX.quantized_matmul, the best way forward would be the config I suggested.

This would allow you to possibly even upstream the model to Bumblebee and we could use EMLX as a pure backend instead of having to drop down to lower levels of abstraction with calling EMLX directly

notactuallytreyanastasio:
@polvalente I would love to do this the right way and get to that point. I just need a small amount of guidance because I am getting to the edge of my knowledge boundaries to make sure all the little parts truly are working together correctly. In the next day or so I will start here and then move through the rest in the same form and see what I can do to get it all in.

polvalente:
@notactuallytreyanastasio Are you on the Elixir or EEF Slack? If you can reach out to me there, we can probably discuss the details more easily!

notactuallytreyanastasio:
@polvalente I am in both but always forget to log into them on new machines. Let me hit you up there.

PR #97, opened 2026-02-22
Use single memcpy for row-contiguous tensors in to_blob

dannote:
`to_blob` previously used a `ContiguousIterator` to copy data element by element, even when the MLX array was row-contiguous in memory. For a 512×1536 f32 tensor (typical embedding batch), that meant 786,432 individual memcpy calls with iterator stepping overhead.

Now checks `t->flags().row_contiguous` and uses a single `memcpy` when possible, falling back to the element-by-element path only for non-contiguous arrays (e.g. transposed views).

### Benchmarks (Apple M5, 24 GB, GPU tensors, post-eval)

| Shape | Before | After | Speedup |
|-------|--------|-------|---------|
| 1×1536 | 18 µs | 12 µs | 1.5× |
| 64×1536 | 219 µs | 12 µs | 18× |
| 512×1536 | 1,680 µs | 105 µs | 16× |
| 1024×1024 | 2,234 µs | 118 µs | 19× |

Benchmark script included in `bench/to_binary_bench.exs`.

All 2115 existing tests pass, plus 4 new tests covering contiguous, non-contiguous, large tensor, and limited reads.


bench/to_binary_bench.exs
```diff
@@ -0,0 +1,38 @@
+# Benchmark: Nx.to_binary for EMLX tensors
```
polvalente: Let's remove this file

polvalente (approved):
Awesome change!

test/emlx/to_binary_test.exs
```diff
@@ -0,0 +1,40 @@
...
+    assert Nx.to_flat_list(t) == for(<<f::float-32-native <- bin>>, do: f)
+  end
+
+  test "to_binary for large contiguous tensor" do
+    t = Nx.iota({64, 1536}, type: :f32, backend: EMLX.Backend)
+    bin = Nx.to_binary(t)
+    assert byte_size(bin) == 64 * 1536 * 4
+
+    # Verify first and last elements
+    <<first::float-32-native, _::binary>> = bin
```
polvalente: You could verify the whole binary by defining the same tensor in Nx.BinaryBackend

polvalente (approved):

PR #116, opened 2026-06-14
Fix int32 overflow in elem_count for tensors with >2.1B elements

edenamram:
## Problem

`elem_count` uses `std::accumulate` with an `int` initial value:

```cpp
uint64_t elem_count(std::vector<int> shape) {
  return std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<>{});
}
```

When the initial accumulator value is `int`, all intermediate multiplications are performed in 32-bit signed arithmetic. For shapes whose element-count product exceeds `INT32_MAX` (~2.1B), the product overflows to a negative `int`, which is then cast to a large `uint64_t` value. The binary-size guard in `from_blob` then always fires:

```
** (RuntimeError) Binary size is too small for the requested shape
```

### Reproduction

`google/gemma-4-2b-it` (and larger Gemma-4 variants) include a weight called `embed_tokens_per_layer` with shape `{262144, 8960}`:

```
262144 × 8960 = 2,348,810,240  >  INT32_MAX (2,147,483,647)
```

This causes `elem_count({262144, 8960})` to return `18,446,744,071,762,394,560` instead of `2,348,810,240`, making it impossible to load the model via `EMLX.Backend`.

## Fix

Seed the accumulator with `uint64_t{1}` and use `std::multiplies<uint64_t>` so every partial product stays in 64-bit arithmetic:

```cpp
uint64_t elem_count(std::vector<int> shape) {
  return std::accumulate(shape.begin(), shape.end(), uint64_t{1}, std::multiplies<uint64_t>{});
}
```

## Context

Discovered while implementing Gemma-4 support in Bumblebee: [elixir-nx/bumblebee#460](https://github.com/elixir-nx/bumblebee/pull/460).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/emlx/c_src/emlx_nif.cpp
+++ b/emlx/c_src/emlx_nif.cpp
@@ -218,7 +218,7 @@ NIF(to_blob) {
 }
 
 uint64_t elem_count(std::vector<int> shape) {
-  return std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<>{});
+  return std::accumulate(shape.begin(), shape.end(), uint64_t{1}, std::multiplies<uint64_t>{});
 }
 
 NIF(from_blob) {
--- a/.github/workflows/emlx.yml
+++ b/.github/workflows/emlx.yml
@@ -72,7 +72,7 @@ jobs:
           mix test --warnings-as-errors
 
   macos:
-    name: macOS (${{ matrix.job.elixir }}, ${{ matrix.job.otp }})
+    name: macOS ${{ matrix.job.gpu && 'gpu' || 'cpu' }} (${{ matrix.job.elixir }}, ${{ matrix.job.otp }})
     runs-on: macos-26
     strategy:
       fail-fast: false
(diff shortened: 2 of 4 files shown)


emlx/c_src/emlx_nif.cpp
```diff
@@ -218,7 +218,7 @@ NIF(to_blob) {
 }
 
 uint64_t elem_count(std::vector<int> shape) {
-  return std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<>{});
+  return std::accumulate(shape.begin(), shape.end(), uint64_t{1}, std::multiplies<uint64_t>{});
```
polvalente, line 221: could you add a regression test that would catch this?

edenamram: i added a test for this. just run :
```
mix test --include large_tensor
```

polvalente (approved):
Thanks for the quick fix!

PR #117, opened 2026-06-18
Avoid host round-trips for same-device backend copies

hfiguera:
## Summary

This adds a same-device fast path for `Nx.backend_copy/3` when copying EMLX tensors back to `EMLX.Backend`.

Previously, same-device EMLX copies went through the generic binary path, which materializes tensor data through the host before rebuilding an EMLX tensor. This already preserved copy independence, but it was unnecessary when the source and target EMLX device were the same.

This PR adds `EMLX.copy/1`, backed by MLX's `mlx::core::copy`, and uses it for same-device EMLX copies. Cross-device copies still use `EMLX.to_device/2`.

It also makes same-device `Nx.backend_transfer/3` a no-op, which matches transfer semantics more directly when the tensor is already on the requested backend/device.

I exposed `EMLX.copy/1` because MLX already exposes `copy` and EMLX exposes similar low-level tensor primitives directly. If you prefer not to expand the public API in this PR, I can keep the copy helper internal to the backend.

## Local check

On an M4 Max, copying a `{2048, 2048}` f32 tensor from EMLX CPU to the same EMLX CPU device preserved values and independent refs while avoiding the host binary round-trip.

Rough local timing for:

```elixir
Nx.backend_copy(t, {EMLX.Backend, device: :cpu})
```

| branch | median |
| --- | ---: |
| `upstream/main` | ~252 us |
| this branch | ~13 us |

For copy followed by materialization with `Nx.to_binary/1`:

| branch | median |
| --- | ---: |
| `upstream/main` | ~264 us |
| this branch | ~40 us |

This is a small local sanity check, not a full benchmark suite.

## Tests

- `mix test test/emlx/nx_test.exs --include metal`

### DIFF
--- a/emlx/c_src/emlx_nif.cpp
+++ b/emlx/c_src/emlx_nif.cpp
@@ -160,6 +160,13 @@ NIF(astype) {
   TENSOR(mlx::core::astype(*t, type, device));
 }
 
+NIF(copy) {
+  TENSOR_PARAM(0, t);
+  DEVICE_PARAM(1, device);
+
+  TENSOR(mlx::core::copy(*t, device));
+}
+
 // Builds the resource binary for `to_blob` in `out_env`. Used by both the
 // legacy direct-call NIF and `command_queue_post_to_blob` (which builds the
 // term inside the worker thread for delivery via enif_send). May throw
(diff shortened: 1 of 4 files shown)


polvalente (approved):
Very nice!

PR #119, opened 2026-06-20
Support dense Qwen3 generation in EMLXAxon

hfiguera:
This adds a dense Qwen3 generation path to EMLXAxon.

The existing Qwen3 path is built around MLX 4 bit checkpoints. This PR adds support for loading standard Hugging Face dense safetensors for Qwen3 and running generation through the native EMLX path.

The main pieces are:

- load dense Qwen3 safetensors into `EMLXAxon.Qwen3.Model.State`
- support greedy dense generation through `EMLXAxon.Qwen3.Generate`
- support `EMLXAxon.TextGeneration.run/4`, `stream/5`, and `serving/3`
- add the native EMLX support needed by the dense Qwen3 path
- add validation around shapes, cache sizes, batch size, and generation options
- add numerical and behavior tests for the dense path

The existing quantized path is still covered. The dense path is separate from the MLX 4 bit loader and is intended for standard Hugging Face checkpoints such as `Qwen/Qwen3-0.6B`.

Tests:

- `cd emlx && mix test test/emlx/fast_test.exs --include metal`
- `cd emlx_axon && mix test test/emlx/qwen3_dense_loader_test.exs test/emlx/qwen3_generate_test.exs test/emlx/text_generation_test.exs --include metal`

Local smoke numbers on an M4 Max with Qwen3 0.6B and 32 generated tokens:

- MLX 4 bit path: around 118 tokens/sec p50
- dense safetensors path: around 244 tokens/sec p50

These are local numbers, but they helped verify that the dense path is operational and not just functionally correct.


emlx/c_src/emlx_fast.cpp
```diff
@@ -640,3 +640,1681 @@ NIF(kv_cache_sdpa_update) {
   CATCH()
 }
 ASYNC_NIF(kv_cache_sdpa_update)
+
+static ERL_NIF_TERM qwen3_error(ErlNifEnv *env, const std::string &message) {
+  return nx::nif::error(env, message.c_str());
+}
```
polvalente: I don't think we need this as a separate function

emlx/c_src/emlx_fast.cpp
```diff
@@ -640,3 +640,1681 @@ NIF(kv_cache_sdpa_update) {
...
+
+static bool qwen3_check_rank(
```
polvalente: How about adding a separate emlx_fast/qwen3.cpp file and its .h companion? I think this would result in cleaner code require fewer qwen3_* prefixed functions

emlx/c_src/emlx_fast.cpp
```diff
@@ -640,3 +640,1681 @@ NIF(kv_cache_sdpa_update) {
...
+  }
+
+  if (!qwen3_check_dim(k_cache, 0, batch, "k_cache", "batch", error) ||
+      !qwen3_check_dim(v_cache, 0, batch, "v_cache", "batch", error) ||
+      !qwen3_check_dim(k_cache, 1, num_kv_heads, "k_cache", "heads", error) ||
+      !qwen3_check_dim(v_cache, 1, num_kv_heads, "v_cache", "heads", error) ||
+      !qwen3_check_dim(k_cache, 3, head_dim, "k_cache", "head_dim", error) ||
+      !qwen3_check_dim(v_cache, 3, head_dim, "v_cache", "head_dim", error)) {
+    return false;
+  }
```
polvalente: This seems like something that should be validated at the Elixir layer

emlx/c_src/emlx_fast.cpp
```diff
@@ -640,3 +640,1681 @@ NIF(kv_cache_sdpa_update) {
...
+//   scale    — float                1/sqrt(head_dim)
+//   head_dim — int                  RoPE dimensions
+//   theta    — float                RoPE base
+//   device   — atom
+//
+// Returns {attn_out, k_upd, v_upd}:
+//   attn_out — {B, T_new, N_q * D}  ready for projection BTH layout
+//   k_upd    — {B, N_kv, T_max, D}
+//   v_upd    — {B, N_kv, T_max, D}
+NIF(qwen3_kv_cache_attention) {
```
polvalente: What does this function do that fast_sdpa or fast_sdpa_masked don't do already?
Maybe we can remove this in favor of composing over those functions in Elixir or extending either of them?

emlx/c_src/emlx_fast.cpp
```diff
@@ -640,3 +640,1681 @@ NIF(kv_cache_sdpa_update) {
...
+    const mlx::core::array &weight,
+    const mlx::core::Device &device) {
+  if (x.ndim() == 3 && x.shape(1) == 1) {
+    auto x_2d = mlx::core::reshape(x, {x.shape(0), x.shape(2)}, device);
+    auto out = mlx::core::matmul(x_2d, weight, device);
+    return mlx::core::reshape(out, {x.shape(0), 1, weight.shape(1)}, device);
+  }
+
+  return mlx::core::matmul(x, weight, device);
+}
```
polvalente: This seems like a function that should be composed in elixir instead of requiring native custom code

emlx/c_src/emlx_nif_shared.hpp
```diff
@@ -12,6 +12,7 @@
 #include <cmath>
 #include <cstring>
 #include <limits>
+#include <memory>
```
polvalente: This should be included in the file that uses it

emlx_axon/lib/emlx_axon/qwen3/attention.ex
```diff
@@ -40,64 +42,159 @@ defmodule EMLXAxon.Qwen3.Attention do
...
+        v_proj,
+        o_proj,
+        q_norm,
+        k_norm,
+        cfg
+      )
+    end
+  end
+
+  defp forward_generic(
```
polvalente: Could we have a better (more specific) name?

emlx_axon/lib/emlx_axon/qwen3/attention.ex
```diff
@@ -40,64 +42,159 @@ defmodule EMLXAxon.Qwen3.Attention do
...
+         k_norm,
+         cfg
+       ) do
     num_heads = cfg.num_attention_heads
     num_kv_heads = cfg.num_key_value_heads
     head_dim = cfg.head_dim
     scale = 1.0 / :math.sqrt(head_dim)
     theta = cfg.rope_theta
 
+    hidden_norm = Layers.rms_norm(hidden, norm, cfg.rms_norm_eps)
```
polvalente, line 98: This looks like a breaking change. Just making note of this so I can check the caller code elsewhere.

emlx_axon/lib/emlx_axon/qwen3/attention.ex
```diff
@@ -40,64 +42,159 @@ defmodule EMLXAxon.Qwen3.Attention do
...
+    # Fused Q/K transpose + RoPE + cache update + SDPA.
     #
     # The NIF move-extracts k_cache / v_cache from their ENIF resources before
     # slice_update, enabling MLX's donation optimisation at eval time: the
     # existing 4 MB Metal buffer is reused in-place — no new allocation needed.
     # The slice and SDPA are fused in the same lazy graph, so they land in one
     # Metal command buffer submission.
     {attn_ref, k_cache_ref, v_cache_ref} =
-      EMLX.kv_cache_sdpa_update(
+      EMLX.qwen3_kv_cache_attention(
```
polvalente, line 126: This supports my general impression that we can avoid having the C++ code in favor of extending this implementation.

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
```diff
@@ -0,0 +1,288 @@
...
+  defp tensor!(params, scope, name) do
+    case params do
+      %{^scope => %{^name => tensor}} ->
+        tensor
+
+      _missing ->
+        nil
+    end
+    |> case do
+      %Nx.Tensor{data: %EMLX.Backend{ref: {:gpu, _ref}}} = tensor -> tensor
```
polvalente:
```suggestion
```

We don't have to check the internal structure of the tensor, backend_transfer will be a noop if on the same device :)

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
```diff
@@ -0,0 +1,288 @@
...
+  defp maybe_cast(tensor, :f16), do: Nx.as_type(tensor, :f16)
+  defp maybe_cast(tensor, :bf16), do: Nx.as_type(tensor, :bf16)
+
+  defp tensor!(params, scope, name) do
+    case params do
+      %{^scope => %{^name => tensor}} ->
+        tensor
+
+      _missing ->
+        nil
```
polvalente: maybe raise here and just do the backend_transfer above, to avoid sequential case

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
```diff
@@ -0,0 +1,288 @@
...
+    end
+  end
+
+  defp eval_state(%State{} = state) do
+    state
+    |> state_tensors()
+    |> Enum.each(fn tensor ->
+      tensor
+      |> EMLX.Backend.from_nx()
+      |> EMLX.eval()
```
polvalente: Why do we want to eval_state here?

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -124,42 +185,175 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+    {rng_key, gpu_key} = advance_rng_for_gpu(sampler, rng_key)
+
+    {next_token_id, kv_new} =
+      forward_sample_token_id(last_id, kv, cur, sampler, temp, top_p, gpu_key, state)
+
+    emit_token(token_callback, next_token_id)
+
+    acc_times =
+      if profile_timing? do
+        te = System.monotonic_time(:millisecond)
```
polvalente: we might want to profile using microsecond

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -124,42 +185,175 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+    )
+  end
+
+  defp decode_tensors(0, acc_tokens, acc_times, kv, cur, rng_key, ctx) do
+    {eos_id, _sampler, _temp, _top_p, _state, _profile_timing?, _token_callback, _chunk_callback} =
+      ctx
+
+    tokens =
+      acc_tokens
+      |> :lists.reverse()
```
polvalente:
```suggestion
      |> Enum.reverse()
```

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -124,42 +185,175 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+      |> tokens_to_host()
+      |> truncate_at_eos(eos_id)
+
+    {:lists.reverse(tokens), :lists.reverse(acc_times), kv, cur, rng_key}
```
polvalente:
```suggestion
    {Enum.reverse(tokens), Enum.reverse(acc_times), kv, cur, rng_key}
```

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -169,19 +363,264 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+    case maybe_flush_chunk(
+           host_tokens,
+           pending_tokens,
+           pending_count,
+           eos_id,
+           chunk_size,
+           chunk_callback
+         ) do
+      {:halt, tokens} ->
+        {:lists.reverse(tokens), :lists.reverse(acc_times), kv, cur, rng_key}
```
polvalente:
```suggestion
        {Enum.reverse(tokens), Enum.reverse(acc_times), kv, cur, rng_key}
```

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -169,19 +363,264 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+  defp forward_sample_token_id(token_id, kv, cur, :greedy, _temp, _top_p, _key, state)
+       when is_integer(token_id) do
+    Model.forward_greedy_decode_token_id(token_id, kv, cur, state)
+  end
+
+  defp forward_sample_token_id(token_id, kv, cur, sampler, temp, top_p, key, state)
+       when is_integer(token_id) do
+    input_ids = Nx.tensor([[token_id]], type: :s64, backend: @cpu_backend)
+    {next_token, kv_new} = forward_sample(input_ids, kv, cur, sampler, temp, top_p, key, state)
+    {Nx.to_number(next_token), kv_new}
```
polvalente, line 626: We should treat Nx.to_number as a costly operation. Is there a way to defer this to the end of the execution pipeline?

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -190,6 +629,88 @@ defmodule EMLXAxon.Qwen3.Generate do
     Sampler.top_p_gpu(logits, gpu_key, temperature: temp)
   end
 
+  defp emit_token(nil, _token_id), do: :ok
+
+  defp emit_token(callback, token_id) when is_function(callback, 1) do
+    callback.(token_id)
+    :ok
+  end
```
polvalente, line 641: Maybe use Nx hooks here?

emlx_axon/lib/emlx_axon/qwen3/generate.ex
```diff
@@ -190,6 +629,88 @@ defmodule EMLXAxon.Qwen3.Generate do
...
+
+  defp eos_seen?(tokens, eos_id), do: Enum.any?(tokens, &eos_token?(&1, eos_id))
+
+  defp eos_token?(token_id, eos_ids) when is_list(eos_ids), do: token_id in eos_ids
+  defp eos_token?(token_id, eos_id), do: token_id == eos_id
+
+  defp tokens_to_host(tokens) do
+    tokens
+    |> Nx.stack()
+    |> Nx.to_flat_list()
```
polvalente:
```suggestion
    Enum.flat_map(tokens, &Nx.to_flat_list/1)
```

maybe this works?

emlx_axon/lib/emlx_axon/qwen3/sampler.ex
```diff
@@ -14,8 +14,11 @@ defmodule EMLXAxon.Qwen3.Sampler do
...
   @doc "Greedy decoding: return the token with the highest logit as a scalar tensor."
   def greedy(logits) do
-    # {1, vocab} → 1-D logits → argmax without trailing squeeze on a {1} intermediate.
-    logits |> Nx.squeeze(axes: [0]) |> Nx.argmax(axis: 0)
+    logits
+    |> EMLX.Backend.from_nx()
+    |> EMLX.argmax(1, false)
+    |> EMLX.Backend.to_nx()
+    |> Nx.squeeze(axes: [0])
   end
```
polvalente: Can you explain the need for this change?

emlx_axon/lib/emlx_axon/text_generation.ex
```diff
@@ -20,52 +20,345 @@ defmodule EMLXAxon.TextGeneration do
...
+    %{text: text} = normalize_input(input)
+
+    encoded =
+      Nx.with_default_backend(Nx.BinaryBackend, fn ->
+        Bumblebee.apply_tokenizer(tokenizer, [text])
+      end)
+
+    input_ids =
+      encoded["input_ids"]
+      |> Nx.backend_transfer({EMLX.Backend, device: :gpu})
```
polvalente: Let's avoid single-pipe altogether, if possible

polvalente (commented):
The PR overall looks good. One general achitectural change I'd like to propose is that it looks to me like all of the native code can be composed in EMLXAxon, in the Elixir layer.

Can we start with that and then optimize with custom C++ if benchmarks support a huge difference?

Additionally, I'd like to ensure we don't have performance regressions. Can we run this [benchmark](https://github.com/elixir-nx/emlx/blob/main/emlx_axon/bench/validate_qwen3.exs) both in `main` and on this branch?

hfiguera:
@polvalente Thanks, that makes sense.

I ran `emlx_axon/bench/validate_qwen3.exs` as requested, using the same local `mlx-community/Qwen3-0.6B-4bit` checkpoint on main and on this branch. These runs were on an Apple M4 Max. I did not see a regression in the existing paths. The base and rewrite paths stayed in the same range, and the native path improved:

- upstream/main: bb base 10.8 tok/s, bb+rewrite 44.7 tok/s, native 75.4 tok/s
- this branch: bb base 10.7 tok/s, bb+rewrite 50.0 tok/s, native 112.1 tok/s

I also looked back through the local history behind this branch, and ran this as a dense Qwen3 matrix instead of relying only on the existing MLX 4bit benchmark.

The dense benchmark uses the standard Hugging Face `Qwen/Qwen3-0.6B` checkpoint from local safetensors, f16, on the EMLX GPU backend, on the same Apple M4 Max. The first dense path was mostly composed in `emlx_axon` using Elixir plus existing `EMLX` pieces, without exported Qwen3 native kernels. That branch records no `qwen3_*` exports from `EMLX` and reaches about 69 tok/s on this benchmark.

Adding the native pieces cumulatively gives:

- native KV attention: about 100 tok/s
- native MLP: about 112 tok/s
- native attention block: about 201 tok/s
- native full layer: about 219 to 224 tok/s
- native all layer greedy path: about 241 tok/s
- native chunked decode: about 256 tok/s

The current PR branch, using the direct `EMLXAxon.TextGeneration` path, measured about 257 tok/s p50 with the same dense model, precision, and GPU settings.

So I agree the Qwen3 user-facing path belongs in `emlx_axon`. The reason I kept the native support in this PR is that the composed dense path is measurable but substantially slower, and the cumulative matrix shows where the main gains come from.

Some of the native steps are enabling work rather than standalone features, so I think the clearest way to review them is against the end to end dense generation path. I’m happy to split this into smaller follow-up PRs, as long as each split stays tied to an `emlx_axon` dense generation benchmark so the performance reason stays visible.

emlx_axon/lib/emlx_axon/qwen3/generate.ex
hfiguera: Yes, it works and makes sense.

emlx_axon/lib/emlx_axon/text_generation.ex
hfiguera: Updated, thanks.

polvalente:
> @polvalente Thanks, that makes sense.
> 
> I ran `emlx_axon/bench/validate_qwen3.exs` as requested, using the same local `mlx-community/Qwen3-0.6B-4bit` checkpoint on main and on this branch. These runs were on an Apple M4 Max. I did not see a regression in the existing paths. The base and rewrite paths stayed in the same range, and the native path improved:
> 
> * upstream/main: bb base 10.8 tok/s, bb+rewrite 44.7 tok/s, native 75.4 tok/s
> * this branch: bb base 10.7 tok/s, bb+rewrite 50.0 tok/s, native 112.1 tok/s
> 
> I also looked back through the local history behind this branch, and ran this as a dense Qwen3 matrix instead of relying only on the existing MLX 4bit benchmark.
> 
> The dense benchmark uses the standard Hugging Face `Qwen/Qwen3-0.6B` checkpoint from local safetensors, f16, on the EMLX GPU backend, on the same Apple M4 Max. The first dense path was mostly composed in `emlx_axon` using Elixir plus existing `EMLX` pieces, without exported Qwen3 native kernels. That branch records no `qwen3_*` exports from `EMLX` and reaches about 69 tok/s on this benchmark.
> 
> Adding the native pieces cumulatively gives:
> 
> * native KV attention: about 100 tok/s
> * native MLP: about 112 tok/s
> * native attention block: about 201 tok/s
> * native full layer: about 219 to 224 tok/s
> * native all layer greedy path: about 241 tok/s
> * native chunked decode: about 256 tok/s
> 
> The current PR branch, using the direct `EMLXAxon.TextGeneration` path, measured about 257 tok/s p50 with the same dense model, precision, and GPU settings.
> 
> So I agree the Qwen3 user-facing path belongs in `emlx_axon`. The reason I kept the native support in this PR is that the composed dense path is measurable but substantially slower, and the cumulative matrix shows where the main gains come from.
> 
> Some of the native steps are enabling work rather than standalone features, so I think the clearest way to review them is against the end to end dense generation path. I’m happy to split this into smaller follow-up PRs, as long as each split stays tied to an `emlx_axon` dense generation benchmark so the performance reason stays visible.

Wow! That's a really substantial difference gain, so it's basically a no-brainer in terms of "yeah, let's keep the native accelerators on". Would you be up for trying to make this more of a plug-in that emlx_axon compiles?
We could have some core C-level abstractions shared in a .h file, such that emlx_axon can rely on this contract by emlx, and provide the same types of native constructs. This would require making the internal tensor representation documented as well, which we can have as a follow-up.

As a possible iteration plan, we can try to keep the native code in EMLX in the PR, but in a very isolated area, such that it's easier to extract out into an emlx_axon plug-in in the next couple PRs. What do you think?

polvalente:
@hfiguera I also messaged you at Elixir Slack if you wanna discuss things there!

emlx_axon/lib/emlx_axon/qwen3/generate.ex
hfiguera: Makes sense, I’ll switch the per-token profiling to microseconds.

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
hfiguera: Makes sense, good catch. I’ll avoid matching on backend internals and let `backend_transfer/2` handle the noop case.

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
hfiguera: Makes sense. Updated.

emlx_axon/lib/emlx_axon/qwen3/generate.ex
hfiguera: I traced this path. In this generator the callback receives a host token id after `Nx.to_number/1` or the native token-id decode path, so there is no tensor expression left to attach an Nx hook to. I think hooks would fit if we refactor this into a `Nx.Serving.streaming(hooks: [:token])` style path, but for the current Elixir generation loop the plain callback seems simpler.

polvalente: I meant that we could attach the hook instead of using Nx.to_number.

Though in hindsight, for EMLX this will probably not make that much of a difference today. We can likely skip this in favor of the wider structural changes this PR will bring

emlx_axon/lib/emlx_axon/qwen3/attention.ex
hfiguera: Good point. Something like `forward_quantized` is more accurate here.

hfiguera: Updated, thanks.

emlx_axon/lib/emlx_axon/qwen3/generate.ex
hfiguera: Got it, let’s leave this as is for now 👍

emlx_axon/lib/emlx_axon/qwen3/generate.ex
hfiguera: Good point. This is the remaining per-token fallback path. The deferred host sync paths already keep token tensors until chunk/end, and the dense greedy streaming path uses the native token-id helper. Avoiding this `Nx.to_number/1` here would be a larger change and probably require carrying token tensors through the per-token loop and emitting through hooks/Serving. Would you be okay with leaving that for a follow-up PR?

polvalente: Yes! I think the most important part fot this PR is laying the foundation for an easier model-specific extension path

emlx_axon/lib/emlx_axon/qwen3/dense_loader.ex
hfiguera: Good question. I added it to materialize the dense safetensors state during load so the first generation call would be less noisy. But thinking more about it, that is probably too opinionated for the loader. Other paths usually make this explicit as preallocation/warmup, so I’ll remove it here and leave warming the state to the caller. We can add an explicit option later in a separate PR if needed.

emlx_axon/lib/emlx_axon/qwen3/sampler.ex
hfiguera: I added this to bypass the generic `Nx.argmax/2` path, which is faster in isolation. But I checked it against the full benchmark and it is not the main reason for the native improvement, and it is mostly a quantized fallback optimization rather than part of the dense path. I’ll revert this to keep the PR focused.

hfiguera:
@polvalente Yes, that direction makes sense to me.

For this PR, I think the safest step is to keep the native code in `emlx` for now, but make the boundary much clearer: isolate the Qwen3 native code from the generic EMLX fast path, keep the user facing API in `emlx_axon`, and avoid spreading model specific code through unrelated native files.

I would prefer not to move this into an `emlx_axon` native plug-in in this same PR, because that also needs a documented C contract for EMLX tensor refs and probably some packaging and build design. But I agree that this PR can be structured so that extraction is straightforward in follow-up PRs.

A concrete plan could be:

1. Keep the current native accelerators in this PR.
2. Move the Qwen3 native pieces into an isolated C++/header area inside `emlx`.
3. Keep the Elixir-facing dense loading and generation path in `emlx_axon`.
4. In a follow-up, document the EMLX tensor/native contract and explore compiling the model-specific native extension from `emlx_axon`.

That would preserve the performance work while making the ownership boundary clearer.

emlx/c_src/emlx_fast.cpp
hfiguera: Agreed. I removed the helper and call `nx::nif::error(...)` directly now.

polvalente:
> @polvalente Yes, that direction makes sense to me.
> 
> For this PR, I think the safest step is to keep the native code in `emlx` for now, but make the boundary much clearer: isolate the Qwen3 native code from the generic EMLX fast path, keep the user facing API in `emlx_axon`, and avoid spreading model specific code through unrelated native files.
> 
> I would prefer not to move this into an `emlx_axon` native plug-in in this same PR, because that also needs a documented C contract for EMLX tensor refs and probably some packaging and build design. But I agree that this PR can be structured so that extraction is straightforward in follow-up PRs.

I agree with this general plan!

> A concrete plan could be:
> 
> 1. Keep the current native accelerators in this PR.
> 2. Move the Qwen3 native pieces into an isolated C++/header area inside `emlx`.
> 3. Keep the Elixir-facing dense loading and generation path in `emlx_axon`.

I think these 3 steps are the cutoff for this PR.

> 4. In a follow-up, document the EMLX tensor/native contract and explore compiling the model-specific native extension from `emlx_axon`.
> 
> That would preserve the performance work while making the ownership boundary clearer.


Perfect. I think with having this native and possibly a separate NIF boundary for Qwen3 while still in EMLX can be left as is, and when we have other specific accelerated models we think about extracting common features.

emlx/c_src/emlx_fast.cpp
hfiguera: Done. I moved the Qwen3 native code to `emlx/c_src/emlx_fast/qwen3.cpp` and `emlx/c_src/emlx_fast/qwen3.hpp`, and kept the NIF API unchanged.

emlx/c_src/emlx_fast.cpp
hfiguera: I agree for cheap public API checks. I kept these in C++ because this path receives raw EMLX tensor refs and this is the last boundary before MLX shape access/cache updates. Duplicating all KV cache shape checks in Elixir would add extra `EMLX.shape/1` calls in the generation hot path.

polvalente: Fine by me!

emlx/c_src/emlx_fast.cpp
hfiguera: This does more than SDPA. It is the Qwen3 attention cache step: transpose from `{B, T, N, D}` to the native attention layout, apply Qwen3 RoPE to Q/K with the cache offset, update the KV cache, slice the valid prefix, build the prefill causal mask when needed, run SDPA, then return flattened `{B, T, N * D}` for the output projection.

This can be composed from smaller EMLX ops, but that was one of the slower paths in the dense benchmark. The native KV attention step moved the dense path from about 69 tok/s to about 100 tok/s.

emlx/c_src/emlx_fast.cpp
hfiguera: That makes sense if this were standalone. It is now private inside `emlx/c_src/emlx_fast/qwen3.cpp` and only used inside the fused Qwen3 kernels.

emlx/c_src/emlx_nif_shared.hpp
hfiguera: You’re right. I moved `<memory>` into `emlx_fast/qwen3.cpp`, where it is used.

emlx_axon/lib/emlx_axon/qwen3/attention.ex
hfiguera: Good catch. The internal caller was updated, so the Qwen3 path in this PR is consistent, but this does change the direct `Attention.forward/11` shape. I made that change so quantized and dense attention both receive the same pre-norm residual input and apply `norm1` inside the attention path.

If you prefer to avoid that public shape change, I can keep a compatibility wrapper for the old arity.

polvalente: I think it's fine if we change things in EMLX Axon. Performance is paramount here I'd say. And the library is still experimental in the sense that this is the only model we support

polvalente (approved):
Awesome contribution! Thanks a ton!!

Reiterating, I really want to connect over Slack so we can discuss possible improvements to the library UX :)

PR #126, opened 2026-07-14
Add compiler support for the EMLXAxon Qwen3 plugin

hfiguera:
## Summary

This moves the Qwen3 native compute code out of EMLX and into a shared library built by `emlx_axon`.

EMLX now provides the generic plugin ABI, registry, compiler lowering, loading, and artifact validation. `emlx_axon` owns the Qwen3 callbacks, schemas, and generation behavior.

The plugin works for both eager execution and EMLX compiled defns. This also removes `EMLX.Native.Qwen3` and the Qwen3 NIF wrappers from EMLX.

## Why

The goal is to keep model implementations in `emlx_axon` without losing fused native operations or defn compatibility.

Qwen3 is the first complete implementation of this path. Other model accelerators should be able to use the same plugin boundary without adding model specific APIs to EMLX.

## Performance

Measured on an Apple M4 Max with 128 GB RAM using `Qwen/Qwen3-0.6B`, f16, greedy generation, 64 generated tokens, and the EMLX GPU backend:

| Mode | Upstream main | This branch | Difference |
| --- | ---: | ---: | ---: |
| End synchronization | 270.49 tok/s | 272.55 tok/s | +0.76% |
| Chunk streaming | 259.25 tok/s | 270.20 tok/s | +4.22% |

Both revisions produced the same token count, finish reason, and decoded text hash.

The real `mlx-community/Qwen3-0.6B-4bit` path also passes its deterministic token and decoded text checks.

## Verification

- `mix format --check-formatted`
- EMLXAxon test suite: 103 passed
- Compiled and eager plugin parity tests
- Real MLX 4-bit inference tests: 2 passed
- Dense f16 end and chunk generation benchmarks
- Plugin loading, lifecycle, build identity, and malformed metadata tests


.github/workflows/emlx_axon.yml
```diff
@@ -25,6 +47,9 @@ jobs:
     runs-on: macos-26
     env:
       MIX_ENV: test
+      # This workflow validates EMLX and EMLXAxon changes together. The Hex
+      # release does not contain branch-only native plugin APIs yet.
```
polvalente:
```suggestion
      # This workflow validates EMLX and EMLXAxon changes together.
```

I think we can leave this indefinitely

emlx/c_src/tools/emlx_build_identity.cpp
```diff
@@ -0,0 +1,551 @@
+#include "emlx_sha256.hpp"
```
polvalente: Can we remove all of this build identity tooling in favor of https://github.com/cocoa-xu/cc_precompiler?
We could have a release workflow that builds the plugin .so for a few common OS/arch combos, and then the fingerprinting of the .so is handled.

The only thing that we wouldn't ensure I think is EMLX ABI compatibility, but I think we can just deal with that if it becomes a problem. Maybe have the header introduce symbol for the version that we can check against.

polvalente: For now, I'm skipping anything related to binary fingerprinting

emlx/c_src/emlx_compiler.cpp
```diff
@@ -2034,6 +2034,145 @@ static const std::unordered_map<std::string, MultiOpFn> multi_op_registry = {
...
+    throw std::runtime_error("emlx::native: missing plugin callback attribute count");
+  const int64_t callback_attr_count =
+      plugin_attr_int(attrs[cursor++], "callback attribute count");
+  if (callback_attr_count < 0 || callback_attr_count > 16384 ||
+      cursor + static_cast<size_t>(callback_attr_count) != attrs.size())
+    throw std::runtime_error("emlx::native: malformed plugin callback attributes");
+  std::vector<int64_t> callback_attrs;
+  callback_attrs.reserve(static_cast<size_t>(callback_attr_count));
+  for (; cursor < attrs.size(); ++cursor)
+    callback_attrs.push_back(plugin_attr_int(attrs[cursor], "callback attribute"));
```
polvalente:
```suggestion
  while(cursor < attrs.size()) {
    callback_attrs.push_back(plugin_attr_int(attrs[cursor++], "callback attribute"));
  }
```

polvalente: I think it's best practice to generally avoid block-less for/while/if

emlx/c_src/emlx_compiler.cpp
```diff
@@ -2034,6 +2034,145 @@ static const std::unordered_map<std::string, MultiOpFn> multi_op_registry = {
...
+        "output", plugin_name, callback_name);
+  }
+  if (instr.operands.size() != expected_operands)
+    throw std::runtime_error("emlx::native: plugin operand count mismatch");
+  if (expected_operands > EMLX_PLUGIN_OPERAND_COUNT_MAX_V1)
+    throw std::runtime_error("emlx::native: plugin operand count exceeds its limit");
+  if (expected_outputs > EMLX_PLUGIN_OUTPUT_COUNT_MAX_V1)
+    throw std::runtime_error("emlx::native: plugin output count exceeds its limit");
+  if (templates.size() != expected_outputs)
+    throw std::runtime_error("emlx::native: plugin output count mismatch");
```
polvalente: add curly brackets here maybe

emlx/c_src/emlx/compiler.cpp
```diff
@@ -2034,6 +2034,145 @@ static const std::unordered_map<std::string, MultiOpFn> multi_op_registry = {
...
+    throw std::runtime_error("emlx::native: plugin callback does not support the worker device");
+
+  const auto stream = emlx::g_current_worker->stream();
+  EMLXPluginExecutionContext execution{&device, &stream};
+  EMLXPluginCall call{{operands.data(), operands.size()},
+                      {instr.plugin_attrs.data(), instr.plugin_attrs.size()},
+                      &execution};
+  std::vector<mlx::core::array> candidates;
+  std::string error;
+  bool ok = false;
```
polvalente: I think it's a bit convoluted to have both ok and error defined here. Maybe `callback.callback` could have error be a std::optional a

polvalente: Or simply `if (error != "")` as the check for error

emlx/c_src/emlx_compiler.cpp
```diff
@@ -2034,6 +2034,145 @@ static const std::unordered_map<std::string, MultiOpFn> multi_op_registry = {
...
+  } catch (const std::exception &exception) {
+    error = exception.what();
+  } catch (...) {
+    error = "unknown plugin callback exception";
+  }
+  if (!ok)
+    throw std::runtime_error(emlx_plugin_callback_failure_error(
+        instr.resolved_plugin.plugin->name, callback.name, error));
+  if (candidates.size() != instr.plugin_outputs.size())
+    throw std::runtime_error("emlx::native: plugin callback returned the wrong output count");
```
polvalente: maybe add expected vs actual in the message

emlx/c_src/emlx_compiler.cpp
```diff
@@ -2034,6 +2034,145 @@ static const std::unordered_map<std::string, MultiOpFn> multi_op_registry = {
...
+  for (size_t i = 0; i < candidates.size(); ++i) {
+    if (candidates[i].shape() != instr.plugin_outputs[i].shape ||
+        candidates[i].dtype() != instr.plugin_outputs[i].dtype)
+      throw std::runtime_error("emlx::native: plugin callback output template mismatch");
```
polvalente: likewise

emlx/c_src/emlx_compiler.cpp
polvalente, line 2311: what makes this not `const` anymore? Does `resolve_plugin_instruction` mutate `instr`?

emlx/c_src/emlx/compiler.cpp
```diff
@@ -2160,9 +2306,9 @@ Expr::~Expr() {
 // rather than inside the interpreter at (first) eval time. `:while` itself
 // is valid despite never appearing in op_registry/multi_op_registry — see
 // interpret_instructions.
-static void validate_instructions(const std::vector<Instruction> &instructions,
+static void validate_instructions(std::vector<Instruction> &instructions,
                                   bool &has_runtime_call) {
-  for (const auto &instr : instructions) {
+  for (auto &instr : instructions) {
```
polvalente, line 2323: What makes this not `const` anymore? Does `resolve_plugin_instruction` mutate the instruction? If it doesn't, let's change the type declaration there instead.

emlx/c_src/emlx_compiler.hpp
polvalente, line 1: Let's have a nested directory structure to avoid names-as-paths:

```
emlx/
  compiler.cpp
  compiler.hpp
  plugin/
    abi.hpp
    registry.hpp
    toolchain.hpp
```

and so on

emlx/c_src/emlx_depfile.cpp
```diff
@@ -0,0 +1,251 @@
+#include "emlx_depfile.hpp"
```
polvalente: Is this also part of the plugin identity toolchain?

emlx/c_src/emlx_nif_lifecycle.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+int emlx_reject_nif_upgrade(ErlNifEnv *env, void **priv_data,
+                            void **old_priv_data,
+                            ERL_NIF_TERM load_info) noexcept {
+  (void)env;
+  (void)priv_data;
+  (void)old_priv_data;
+  (void)load_info;
+  return -1;
```
polvalente: we can return success here and allow `recompile` that doesn't change the native code

polvalente (changes requested):
Thanks for the PR. I think this is going in the right direction.
However, we should remove all of the fingerprinting/image/id tooling from this PR.
We can maybe pull-in cc_precompiled to help with securing the immediate surface.
The plugin design seems good, but I'm having trouble decoupling it from the fingerprinting tooling

emlx/c_src/emlx_compiler.cpp
hfiguera: Good point

emlx/c_src/emlx_compiler.cpp
hfiguera: Sure.

emlx/c_src/emlx/compiler.cpp
hfiguera: Good point. What about having the callback return `std::optional<std::string>`? `nullopt` would mean success and a value would be the error. Is that what you had in mind?

emlx/c_src/emlx_compiler.cpp
hfiguera: Good idea 👍

emlx/c_src/emlx/compiler.cpp
hfiguera: It does mutate the instruction by storing the resolved callback, attributes, and output templates. I’ll rename the function to make that clearer.

emlx/c_src/emlx_nif_lifecycle.cpp
hfiguera: Good catch! Yeah, this is stricter than it needs to be.

emlx/c_src/emlx_compiler.hpp
hfiguera: ok

emlx/c_src/emlx_depfile.cpp
hfiguera: Yes. It verifies that EMLX and its plugins were built against compatible MLX headers, so I moved it under `emlx/plugin/` for the moment.

emlx/c_src/tools/emlx_build_identity.cpp
hfiguera: It makes sense. I’ll remove the custom identity tooling and keep local source compilation in this PR. I think `cc_precompiler` fits better in a follow-up once the plugin ABI is settled. Does that work for you?

polvalente: Yes, I think that's the best approach!

emlx/c_src/emlx/compiler.cpp
polvalente: Yeah! I like that approach.

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <type_traits>
+#include <vector>
+
+#if defined(__GNUC__) || defined(__clang__)
+#define EMLX_PLUGIN_EXPORT __attribute__((visibility("default")))
+#else
+#define EMLX_PLUGIN_EXPORT
+#endif
+
+inline constexpr uint64_t EMLX_PLUGIN_MAGIC_V1 = 0x454D4C58504C4731ULL;
```
polvalente: Can we document where does this come from and how one would update these values when updating the abi?

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+inline constexpr uint32_t EMLX_PLUGIN_ABI_V1 = 1;
+inline constexpr uint32_t EMLX_PLUGIN_DEVICE_CPU_V1 = 1U << 0;
+inline constexpr uint32_t EMLX_PLUGIN_DEVICE_GPU_METAL_V1 = 1U << 1;
+inline constexpr uint32_t EMLX_PLUGIN_DEVICE_KNOWN_V1 =
+    EMLX_PLUGIN_DEVICE_CPU_V1 | EMLX_PLUGIN_DEVICE_GPU_METAL_V1;
```
polvalente: aren't there enums in mlx that we can just expose accordingly?

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+inline constexpr uint32_t EMLX_PLUGIN_OPERAND_COUNT_MAX_V1 = 8192;
+inline constexpr uint32_t EMLX_PLUGIN_OUTPUT_COUNT_MAX_V1 = 1024;
```
polvalente: Why are input and output counts limited?

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+// invocation that receives them. Neither side may retain borrowed pointers.
+template <typename T> struct EMLXPluginView {
+  const T *data;
+  uint64_t size;
+};
+
+struct EMLXPluginStringView {
+  const char *data;
+  uint64_t size;
+};
```
polvalente: Isn't this the same as EMLXPluginView<char>?

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+inline constexpr uint32_t EMLX_PLUGIN_DEVICE_KNOWN_V1 =
+    EMLX_PLUGIN_DEVICE_CPU_V1 | EMLX_PLUGIN_DEVICE_GPU_METAL_V1;
+inline constexpr uint32_t EMLX_PLUGIN_OPERAND_COUNT_MAX_V1 = 8192;
+inline constexpr uint32_t EMLX_PLUGIN_OUTPUT_COUNT_MAX_V1 = 1024;
+
+// Plugin-owned views are borrowed by EMLX only while loading the descriptor.
+// Call-owned views remain valid only for the count-policy or callback
+// invocation that receives them. Neither side may retain borrowed pointers.
+template <typename T> struct EMLXPluginView {
+  const T *data;
```
polvalente: maybe we want a `std::shared_ptr` or `std::weak_ptr` instead of a raw pointer

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+  uint64_t size;
+};
+
+using EMLXPluginArrayView = EMLXPluginView<mlx::core::array>;
+using EMLXPluginInt64View = EMLXPluginView<int64_t>;
+
+struct EMLXPluginExecutionContext {
+  const mlx::core::Device *device;
+  const mlx::core::Stream *stream;
+};
```
polvalente: Let's use smart pointers here too

emlx/c_src/emlx/plugin/abi.hpp
polvalente: I think this ABI is C++ and not C. Given that, how about we use namespace emlx::plugin to remove some of the variable and type namespacing?

i.e. EMLX_PLUGIN_ABI_V1 -> emlx::plugin::abi_v1, EMLXPluginView -> emlx::plugin::view_t

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+struct EMLXPluginCall {
+  EMLXPluginArrayView operands;
+  EMLXPluginInt64View attrs;
+  const EMLXPluginExecutionContext *execution;
```
polvalente: smart pointer here too?

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+    throw std::runtime_error("plugin descriptor is incompatible with EMLX");
+
+  const std::string descriptor_name =
+      copy_required_string(descriptor.name, kNameMax, "name");
+  if (!valid_name(descriptor_name) || descriptor_name != requested_name)
+    throw std::runtime_error("plugin descriptor name does not match requested name");
+
+  if (descriptor.callback_count > kCallbacksMax ||
+      (descriptor.callback_count > 0 && !descriptor.callbacks))
+    throw std::runtime_error("plugin callback table is invalid");
```
polvalente: We should have better error messages in general. Some of the validation here and around here can fail for multiple reasons for the same error message

polvalente: Also, styleguide: avoid ifs without {}

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+    callback->schema_version = source.schema_version;
+    callback->attr_schema_version = source.attr_schema_version;
+    callback->operand_count = source.operand_count;
+    callback->operand_count_from_attrs = source.operand_count_from_attrs;
+    callback->output_count = source.output_count;
+    callback->output_count_from_attrs = source.output_count_from_attrs;
+    callback->device_capabilities = source.device_capabilities;
+    callback->callback = source.callback;
+    if (!loaded->callbacks.emplace(callback->name, callback).second)
+      throw std::runtime_error("plugin callback names must be unique");
```
polvalente: Most of this could be extracted into a copy constructor for descriptor I think

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  {
+    std::unique_lock lock(g_plugin_mutex);
```
polvalente: Do we want a lock guard instead of a unique_lock? Seems more well suited given that we open a scope to handle the unlock implicitly only

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+    auto [it, inserted] = g_plugins.emplace(requested_name, loaded);
+    if (!inserted) {
+      if (it->second->canonical_path == resolved_path)
+        return it->second;
+      throw std::runtime_error("plugin registration conflicts with an accepted plugin");
+    }
+  }
+  handle.release();
```
polvalente: Do we control the `handle` data structure? Can we make this part of the destructor?

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+  if (error)
+    throw std::runtime_error(
+        emlx_plugin_callback_failure_error(
+            plugin_name, callback_name, *error, callback_error_max));
+  if (outputs.size() != expected_outputs)
+    throw std::runtime_error("plugin callback returned the wrong output count");
+  return outputs;
+}
+
+#if !defined(EMLX_PLUGIN_REGISTRY_TESTING)
```
polvalente: #ifndef?

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,428 @@
...
+    load_generic_candidate(name, path);
+    return nx::nif::ok(env);
+  } catch (const std::bad_alloc &) {
+    return nx::nif::error(env, "plugin loader allocation failed");
+  } catch (const std::exception &error) {
+    return nx::nif::error(env, bounded_error(error.what()).c_str());
+  } catch (...) {
+    return nx::nif::error(env, "internal plugin loader error");
+  }
+}
```
polvalente, line 501: I'd define this with `fine` instead. If not, let's validate argc

emlx/lib/emlx/native/expr.ex
```diff
@@ -395,13 +395,21 @@ defmodule EMLX.Native.Expr do
   end
 
   defp emit_metadata_instr(id, opcode, operands, attrs, state) do
-    ref = make_ref()
     operand_refs = Enum.map(operands, &Map.fetch!(state.node_to_ref, &1.data.id))
 
+    refs =
+      case {opcode, Enum.at(attrs, 5)} do
```
polvalente: let's destructure `attrs` with pattern matching instead, so we can `_name` the  other leading entries too

emlx/Makefile
```diff
@@ -7,7 +7,10 @@ esc = $(subst $(space),\$(space),$(1))
 
 # Private configuration
 PRIV_DIR = $(MIX_APP_PATH)/priv
-BUILD_DIR = $(EMLX_CACHE_DIR)/emlx-$(EMLX_VERSION)-mlx-$(MLX_VERSION)$(MLX_VARIANT)/objs
+ifeq ($(strip $(EMLX_SOURCE_CACHE_ID)),)
+$(error EMLX_SOURCE_CACHE_ID is required)
+endif
+BUILD_DIR = $(EMLX_CACHE_DIR)/emlx-$(EMLX_VERSION)-mlx-$(MLX_VERSION)$(MLX_VARIANT)-source-$(EMLX_SOURCE_CACHE_ID)/objs
```
polvalente: Let's revert this? I don't think we gain anything by using `EMLX_SOURCE_CACHE_ID`. Please correct me if I'm wrong

emlx/mix.exs
```diff
@@ -63,6 +64,14 @@ defmodule EMLX.MixProject do
     [preferred_envs: [docs: :docs, "hex.publish": :docs]]
   end
 
+  @doc false
+  def source_cache_id(source_root) do
+    source_root
+    |> Path.expand()
+    |> then(&:crypto.hash(:sha256, &1))
+    |> Base.encode16(case: :lower)
+  end
```
polvalente: This is basically re-encoding the invariant directory path

emlx_axon/c_src/qwen3_plugin.cpp
```diff
@@ -1,16 +1,65 @@
-#include "qwen3_plugin_abi.hpp"
+#include "emlx/plugin/abi.hpp"
 
+#include <cstring>
+#include <climits>
 #include <limits>
 #include <optional>
 #include <sstream>
 
+namespace emlx_qwen3_plugin {
```
polvalente:
```suggestion
namespace emlx_axon::plugin::qwen3 {
```

how about we follow this namespacing standard?

polvalente (commented):
I think we're almost there!

Most of my reviews are either stylistic or related to code organization. General structure is looking great!

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Good point. I moved the C++ ABI types and constants under `emlx::plugin` and documented how the magic and ABI version should evolve. The discovery symbol stays extern "C" so it remains stable for `dlsym`.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: You’re right. I replaced the custom device flags with a list of MLX DeviceType values and now use the MLX enum directly during plugin loading and dispatch.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Good question. Those were defensive caps, but the values were arbitrary. I removed them and now validate operand counts against the actual operands and output counts against the encoded templates before allocation. The model specific limits stay in the plugin.

emlx/c_src/emlx/plugin/abi.hpp
polvalente: @hfiguera still pending an answere here

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,153 @@
...
+// invocation that receives them. Neither side may retain borrowed pointers.
+template <typename T> struct view_t {
+  const T *data;
+  uint64_t size;
+};
+
+struct string_view_t {
+  const char *data;
+  uint64_t size;
+};
```
polvalente: Given that we're in C++ land, we should either make this an instance of view_t, or just pass std::string (or references to them) around

polvalente: I particularly lean into using std::string

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+using array_view_t = view_t<mlx::core::array>;
+using int64_view_t = view_t<int64_t>;
```
polvalente: do we really need to pass int64 as a view? the pointer has the same size as copying the int, so the view occupies 2x the size plus change

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+struct CandidateHandle {
+  void *value = nullptr;
+  ~CandidateHandle() {
+    if (value)
+      dlclose(value);
+  }
+  void release() { value = nullptr; }
+};
+
```
polvalente: nice!

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+  return std::all_of(value.begin(), value.end(), [](unsigned char c) {
+    return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') ||
+           (c >= '0' && c <= '9') || c == '_' || c == '.' || c == '-';
+  });
+}
+
+bool valid_device_type(emlx::plugin::device_type_t device_type) {
+  return device_type == mlx::core::Device::DeviceType::cpu ||
+         device_type == mlx::core::Device::DeviceType::gpu;
+}
```
polvalente: nice!

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+bool valid_utf8(const std::string &value) {
```
polvalente: Why are we validating utf8?

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: You’re right. I replaced it with `using string_view_t = view_t<char>`.

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+                                 const char *field) {
+  if (!view.data || view.size == 0 || view.size > limit ||
+      view.size > std::numeric_limits<size_t>::max())
+    throw std::runtime_error(std::string("plugin field ") + field +
+                             " is missing or exceeds its limit");
+  return std::string(view.data, static_cast<size_t>(view.size));
+}
+
+std::string canonical_path(const std::string &path) {
+  char *resolved = realpath(path.c_str(), nullptr);
```
polvalente: if you use std::unique_ptr here you don't have to worry about freeing. For example, there's a missing free before the throw

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+  if (!handle.value) {
+    const char *detail = dlerror();
+    throw std::runtime_error(std::string("failed to load plugin: ") +
+                             (detail ? detail : "unknown loader error"));
+  }
+
+  dlerror();
+  auto discovery = reinterpret_cast<emlx::plugin::discovery_v1_fn_t>(
+      dlsym(handle.value, "emlx_plugin_descriptor_v1"));
+  const char *symbol_error = dlerror();
```
polvalente: Same as above, symbol_error is leaking

emlx/c_src/emlx/plugin/registry.cpp
polvalente: This is still relevant, especially in this file

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+                    callback->supported_devices.end(),
+                    device_type) != callback->supported_devices.end()) {
+        throw std::runtime_error("plugin callback descriptor is invalid");
+      }
+      callback->supported_devices.push_back(device_type);
+    }
+    callback->callback = source.callback;
+    if (!loaded->callbacks.emplace(callback->name, callback).second)
+      throw std::runtime_error("plugin callback names must be unique");
+  }
```
polvalente: I think we're still missing a copy constructor to absorb all of this

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: I think the raw pointer fits better here. The data is borrowed and stays alive for the whole callback, so a `shared_ptr` would add ownership we don’t need. Does that make sense to you?

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,450 @@
...
+call_plugin_impl(ErlNifEnv *env, std::string plugin_name,
+                 std::string callback_name, std::vector<TensorArg> operands,
+                 std::vector<int64_t> attrs, mlx::core::Device device) {
+  (void)env;
+  std::vector<mlx::core::array> arrays;
+  arrays.reserve(operands.size());
+  for (const auto &operand : operands)
+    arrays.push_back(*operand);
+  return emlx_invoke_plugin_callback(plugin_name, callback_name, arrays, attrs,
+                                     device);
```
polvalente: should we make emlx_invoke_plugin_callback accept an array of mx::array pointers? This would avoid having to allocate arrays from operands

emlx/c_src/emlx/plugin/registry.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <unordered_map>
+#include <vector>
+
+// Loads and validates the generic plugin descriptor, then keeps the accepted
+// shared object alive for the VM lifetime.
+ERL_NIF_TERM load_plugin(ErlNifEnv *, int, const ERL_NIF_TERM[]);
+ERL_NIF_TERM call_plugin(ErlNifEnv *, int, const ERL_NIF_TERM[]);
+ERL_NIF_TERM call_plugin_async(ErlNifEnv *, int, const ERL_NIF_TERM[]);
+
+struct EMLXLoadedPluginCallback {
```
polvalente: I think we still should change this file (and the hpp) into the namespaced standard as was done with the abi.hpp

polvalente (commented):
Some more comments! I started reviewing before I realized you weren't done yet!

I'll wait for a explicit review request before the next round!

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Good point. Since `Device` and `Stream` are small value types, would you be okay with storing them directly in `call_t` instead of using smart pointers? That avoids the lifetime issue without adding allocations.

emlx/c_src/emlx/plugin/abi.hpp
polvalente: I think the assumption is incorrect in that we do need the ownership because we are reading the data. weak_ptr I think only works when viewing shared_ptr, so shared_ptr seems to be a better fit.
My main concern with this is that we could easily leak memory by using raw pointers

hfiguera: Maybe I’m missing something, but the data stays alive for the whole callback. Are plugins expected to keep it after returning? If so, `shared_ptr` makes sense.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: You’re right, I hadn’t gotten to that part yet. I’ll improve the validation messages and add braces throughout this file.

emlx/c_src/emlx/plugin/abi.hpp
polvalente: I think so! They are basically enums, right? I'm ok with pass by value in exchange for better safety

emlx/c_src/emlx/plugin/abi.hpp
polvalente: We're dealing with a generic view holder here, so we should make it as resilient as possible.
In the current implementation, the worst that happens is that we add a tiny overhead for shared_ptr management and refcounting if we move to shared_ptr. If we don't do it, there is a slight possibility of us running into a read-after-free because the pointer is being constructed on the fly.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Updated. Thanks.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: That’s fair. I’ll switch the generic view to shared ownership and make sure the `shared_ptr` owns the backing storage, not just the pointer. Is that what you had in mind?

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Exactly. `Device` is a type and index, and `Stream` is an index plus a `Device`. I’ll store both directly in `call_t`.

hfiguera: Changed 👍

emlx/c_src/emlx/plugin/abi.hpp
polvalente: Yeah, something along those lines!

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Good idea. I’ll move the descriptor validation and copying into an `EMLXLoadedPluginCallback` constructor.

hfiguera: Done.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Yes, lock_guard is a better fit here. I’ll change it.

hfiguera: Changed 👍

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Yes, we control it. I will move the handle ownership into `EMLXLoadedPlugin` so its destructor handles `dlclose` and we can remove the explicit `release()`

hfiguera: Moved 👍

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Make sense. I did the change.

emlx/lib/emlx/native/expr.ex
hfiguera: Nice one! I did the change.

emlx/Makefile
hfiguera: It prevents stale objects from being shared across worktrees, but it’s outside this PR’s scope. I’ll revert it.

hfiguera: Reverted 👍

emlx/mix.exs
hfiguera: Right, this was removed along with the source cache ID.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Ok, I changed it. Now, the views share ownership of their backing storage, and also I added a test that retains a view across calls.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: I already addressed it. The views use shared ownership, and the device and stream are stored by value.

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: Make sense. Plugin and callback names now use `std::string`

emlx/c_src/emlx/plugin/abi.hpp
hfiguera: This is a variable length attribute list, not a single integer. Some Qwen callbacks pass many values here. Would replacing the view with `std::vector<int64_t>` make sense to you?

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: I don't remember, but it is not needed. So, I removed it. Good catch!

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: I got it. I did the change 👍

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: I may be missing something, but this case is different. `dlerror()` returns loader owned memory, so we should not free it. Which part do you think is leaking here?

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Already addressed

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: This made sense before the ownership change. The array copies keep retained views alive, while raw pointers wouldn’t. I think we should keep this as is. Does that make sense?

emlx/c_src/emlx/plugin/registry.hpp
hfiguera: Done. The registry cpp and hpp now follow the `emlx::plugin` namespace and naming style.

emlx/c_src/emlx/plugin/abi.hpp
polvalente: Yeap!

emlx/c_src/emlx/plugin/registry.cpp
polvalente: before each throw you are missing free for each of the pointers. you're only freeing them in the happy path 

this is exactly what smart pointers solve and why I was so adamant to use them in that other section of the code

emlx/c_src/emlx/plugin/registry.cpp
polvalente: I'm not sure, I will have to check this in more depth

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,513 @@
...
+      }
+      throw std::runtime_error(
+          "plugin registration conflicts with an accepted plugin");
+    }
+  }
+
+  shared_object_handle_t handle{
+      dlopen(resolved_path.c_str(), RTLD_NOW | RTLD_LOCAL)};
+  if (!handle) {
+    const char *detail = dlerror();
```
polvalente: Just marking that this should be a std::string or some smart pointer

emlx/c_src/emlx/plugin/registry.cpp
```diff
@@ -0,0 +1,513 @@
...
+          alignof(descriptor_t) !=
+      0) {
+    throw std::runtime_error("plugin descriptor pointer is not aligned");
+  }
+
+  const auto &descriptor =
+      *static_cast<const descriptor_t *>(bootstrap.descriptor);
+  if (descriptor.descriptor_size != sizeof(descriptor_t)) {
+    throw std::runtime_error("plugin descriptor size does not match ABI v1");
+  }
```
polvalente: I believe this should be above the static_cast

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,148 @@
...
+  if (size == 0) {
+    return {};
+  }
+  if (!values) {
+    throw std::invalid_argument("plugin view data cannot be null");
+  }
+  return make_view(std::vector<T>(values, values + size));
+}
+
+template <typename T, size_t N>
```
polvalente: style: the template declaration should be above the definition:

```
template <typename T>
view_t<T> make_view...
```

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,148 @@
...
+view_t<T> make_view(const T (&values)[N]) {
+  return make_view(values, N);
+}
+
+using array_view_t = view_t<mlx::core::array>;
+using int64_view_t = view_t<int64_t>;
+using device_type_t = mlx::core::Device::DeviceType;
+using device_view_t = view_t<device_type_t>;
```
polvalente: do we still need these?

emlx/c_src/emlx/plugin/abi.hpp
```diff
@@ -0,0 +1,148 @@
...
+  }
+  return make_view(std::vector<T>(values, values + size));
+}
+
+template <typename T, size_t N>
+view_t<T> make_view(const T (&values)[N]) {
+  return make_view(values, N);
+}
+
+using array_view_t = view_t<mlx::core::array>;
```
polvalente: Here's something to consider:

I looked through the PR and only found these uses of make_view:

```
emlx::plugin::make_view(std::move(operands)) -> view of std::vector
emlx::plugin::make_view(instr.plugin_attrs) -> view of std::vector
auto operand_view = make_view(std::move(operands)); -> view of std::vector
auto attr_view = make_view(std::move(attrs)); -> view of std::vector
supported_devices(emlx::plugin::make_view(kSupportedDeviceTypes)) -> view of an array (could be rewritten as std::vector)
```

all of them either are views of vectors or the single view of array.
I have a hunch that all of the views can be rewritten as std::vector directly, passed by reference where pertinent.

polvalente: I'm working on this, will push shortly

polvalente: Pushed!

polvalente (commented):
I went through the unresolved comments and also updated the code to use std::vector instead of view_t. I think there are very few changes needed before this is ready to merge!

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: I think we were referring to different pointers. `dlerror()` returns loader owned memory, while the `dlopen()` handle is wrapped in an object that automatically calls `dlclose()` on every error path. On success, ownership moves into `loaded_plugin_t`. If I’m missing the pointer you mean, feel free to push the change you have in mind.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Agreed. Since `dlerror()` owns the memory, I will copy it into a `std::string`

hfiguera: Done 👍

emlx/c_src/emlx/plugin/registry.cpp
polvalente: I'm referring to the dlerror const char pointer. The memory is owned by us by the point that variable is not nullptr anymore and therefore must be freed. The const qualifier is just a hint that we won't write to it.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: Good point. The bootstrap already validates the descriptor size before the cast, so I removed the redundant size field from the descriptor itself and kept the bootstrap check as the authoritative one.

emlx/c_src/emlx/plugin/registry.cpp
hfiguera: I checked the `dlerror()` contract. The returned memory should not be freed and may be invalidated by the next `dlerror()` call. Would copying it immediately into an optional `std::string`, like the other error path, address your concern?

polvalente: yes!

hfiguera: Done.

polvalente (approved):

polvalente:
I'm merging this so we can move on to adding more plugins!

polvalente:
Screenshot 2026-07-16 at 23 13 45
Just dropping this here after the merge :D

PR #128, opened 2026-07-17
Add native dense Llama generation to EMLXAxon

hfiguera:
Adds native dense Llama generation to EMLXAxon using the generic plugin path introduced for Qwen3.

The initial target is `meta-llama/Llama-3.2-1B`. The implementation includes:

- loading dense Bumblebee parameters or Hugging Face safetensors
- Llama 3 RoPE scaling and grouped query attention
- native layer, greedy forward, and chunked decode callbacks
- host synchronization after each token, at the end, or in chunks
- `EMLXAxon.TextGeneration` support for Llama states
- numerical reference, compiler lowering, cache, generation, and serving tests
- a benchmark comparing stock Bumblebee, the EMLXAxon rewrite, and the native plugin

The Llama C++ implementation lives in `emlx_axon/c_src/llama_plugin.cpp`. EMLX remains responsible for the generic plugin ABI, registry, and compiler integration.

## Performance

Measured on an Apple M4 Max with 128 GB RAM using `meta-llama/Llama-3.2-1B`, f16 weights, greedy generation, and the EMLX GPU backend.

Results are p50 after one warmup and three measured requests. Each request used a 128-token input and generated exactly 32 tokens.

| Path | TTFT | Request duration | Tokens/sec |
| --- | ---: | ---: | ---: |
| Bumblebee + EMLX | 93.4 ms | 366.3 ms | 87.4 |
| EMLXAxon rewrite | 52.7 ms | 268.4 ms | 119.2 |
| Native Llama plugin | 12.0 ms | 187.4 ms | 170.7 |

The native plugin is about 1.95x faster than stock Bumblebee with EMLX and 1.43x faster than the EMLXAxon rewrite. A second run measured 171.8 tok/s with 11.8 ms TTFT.

All measured requests generated 32 tokens and finished at the configured length limit.

## Verification

- `mix format --check-formatted`
- `mix compile --warnings-as-errors`
- `mix xref graph --format cycles`
- `EMLX_TEST_DEFAULT_GPU=true mix test`
- `mix run bench/validate_llama_dense.exs`

The tests cover numerical agreement with Nx references, grouped query attention, nonzero cache offsets, cache isolation, compiler lowering, text generation, plugin loading, and application restarts.


polvalente:
Tested locally that ollama and the native impl. both reach the same tok/s ballpark if using the f16 version of the weights!

I also got the same ballpark of performance as the bench above.

polvalente (approved):

PR #129, opened 2026-07-21
fix: update to Nx 0.13, replace :token/:attach_token with :io_call

Ljzn:
### Problem

EMLX 0.4.0 depends on `nx ~> 0.12` and references `Nx.Defn.Token`, which was
removed in Nx 0.13. The `:token` and `:attach_token` expression ops were
replaced by a single `:io_call` op.

Compiling EMLX with Nx 0.13 gives:

```
** (CompileError) cannot expand Nx.Defn.Token, module not available
```

### Changes

**`mix.exs`**: Updated version constraint from `~> 0.12` to `~> 0.13`.

**`lib/emlx/native/expr.ex`** (3 changes):

1. **`:token` → `:io_call`** (`expand_node`): Replaced the `:token` match
   (which iterated `%Nx.Defn.Token{hooks: hooks}` and fired each hook's
   callback) with a match on `:io_call`. The new clause extracts the callback
   from `callback_spec` (which can be `{:fn, fun}`, `{:named, name, nil}`, or
   `{:named, name, callback}`) and calls `emit_hook_runtime_call` accordingly.

2. **Removed `:attach_token`** (`expand_node`): In Nx 0.13, `:io_call`'s
   output is the hooked value directly — there is no separate attach step.
   The old `hooked_value_refs` lookup logic is replaced by `set_io_call_ref/3`,
   which maps the `:io_call` node's id to the post-hook MLX ref.

3. **`while_scope_contains_hook?`**: Updated the pattern match from
   `:token` + `Nx.Defn.Token` to `:io_call` + `callback_spec`.

### Testing

- All existing EMLX tests pass (when run against Nx 0.13).
- `Nx.top_k` benchmark runs correctly on EMLX (`Nx.Defn.jit` + GPU).
- `io_call` in `defn` context correctly preserves the side-effect semantics
  (callback fires, output value unchanged).
- Chain of multiple `io_call` nodes, name-only hooks (`:named`),
  and function hooks (`:fn`) all verified.

### Code Review Notes

The `set_io_call_ref/3` helper is placed after the fallback `expand_node/2`
clause to avoid breaking Elixir's function clause grouping rules.

---

*Assisted by DeepSeek V4 Flash*

### DIFF
--- a/emlx/lib/emlx/native/expr.ex
+++ b/emlx/lib/emlx/native/expr.ex
@@ -2011,7 +2011,8 @@ defmodule EMLX.Native.Expr do
   # `EMLX.Native.Expr.t/0`'s `keepalive_refs` doc for where the final tip of
   # this chain ends up.
   defp expand_node(
-         %T{data: %Nx.Defn.Expr{id: id, op: :token, args: [%Nx.Defn.Token{hooks: hooks}]}},
+         %T{data: %Nx.Defn.Expr{id: id, op: :io_call,
+              args: [tensor_expr, callback_spec, _template, _ref]}},
          state
        ) do
     unless MapSet.member?(state.top_scope_ids, id) do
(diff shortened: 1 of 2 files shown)


polvalente (approved):

polvalente:
https://github.com/elixir-nx/emlx/releases/tag/v0.4.1 new tag released with an alternative hotfix for making the nx version more strict

polvalente (approved):
We need a separate fix which is to make :io_call a first-class citizen of the compiler just like :runtime_call is

polvalente:
Closing this PR for now. EMLX 0.4.1 has the correct fix, and we'll need to wait until Bumblebee is released to update EMLXAxon.

For what it's worth, the hard breaking change is not in Nx 0.13, but in Nx main

PR #137, opened 2026-08-20
Cache jit closures instead of re-tracing on every call

### LINKED ISSUE #136: Nx.Defn.jit re-traces the function body on every call
### What happens

`Nx.Defn.jit/2` re-traces the function body on every call. The compiled program
is cached and reused; what is not cached is the trace used to find it.

`__jit__/5` forwards to `__compile__/4`, which drops the `key` Nx hands it and
calls `native_compile/3`:

```elixir
def __jit__(key, vars, fun, args_list, opts) do
  __compile__(key, vars, fun, opts).(args_list)
end

def __compile__(_key, vars, fun, opts) do   # key is discarded
  ...
  wrap_with_queue(queue, native_compile(vars, fun, device))
end

defp native_compile(vars, fun, device) do
  output_expr = fun.(vars)                  # full trace, every call
  ...
  base_key = dispatch_key(output_expr, num_inputs, effective_device)
  {resource, runtime_calls} = get_or_compile_program(base_key, ...)
```

So each call pays for building the expression and then walking it structurally,
only to look up a program that is already in the ETS table.

`Nx.Defn.Compiler`'s own documentation describes `key` as the thing to cache on:

> It receives an opaque `key` used for caching […] Note the `key` does not
> include the `vars` in its cache. Therefore, if you want to cache the result of
> `fun.(vars)`, you likely want to include the vars in the cache key.

### Measurements

Attached script, `{:emlx, "~> 0.1"}` only. A chain of `depth` cheap ops, so the
timing reflects graph traversal rather than arithmetic:

```
  depth   jit      compile   difference
     32    0.59 ms     0.38 ms      0.21 ms
    128    1.01 ms     0.40 ms      0.61 ms
    512    3.12 ms     0.84 ms      2.29 ms
   1024    5.89 ms     1.40 ms      4.48 ms
```

The gap grows linearly with the graph — roughly 4.4 µs per node — and at depth
1024 `jit` is 4.2× slower than `compile`.

On a real workload (a 40-layer Qwen3.6-35B-A3B MoE decode step, ~4300 nodes) the
same difference is 8.35 ms per token out of 89, measured against
`Nx.Defn.compile/3` with identical arguments:

```
  8 layers    jit 20.14 ms   compile 18.25 ms   (9%)
 40 layers    jit 89.42 ms   compile 81.07 ms   (9%)
```

An earlier attempt at a reproduction used 64×64 matmuls and showed nothing —
the arithmetic there dominates and hides the traversal. Worth noting for anyone
trying to confirm this.

### Why it matters

`Nx.Defn.jit/2` is the documented entry point and the one most code reaches for.
For anything called in a loop — token-by-token decoding, a training step — the
per-call re-trace is paid on every iteration and scales with model size, which
is the opposite of what a cache is supposed to do. `Nx.Defn.compile/3` avoids
it, but only if the caller knows to reach for it and can supply templates.

### Environment

- emlx 0.4.1 (Hex release, unmodified)
- nx 0.13.1
- libmlx 0.31.2, arm64-apple-darwin
- Elixir 1.20.3, Erlang/OTP 29
- macOS 26.5, Apple Silicon (M3 Pro)

### Possible direction

Cache the closure `__compile__/4` returns, keyed by the `key` Nx supplies plus
the shapes and types of `vars` (and the compiler options that affect lowering,
such as `:device` and `:command_queue`). On a hit, `fun.(vars)` never runs.

<details>
<summary>jit_retrace_repro.exs</summary>

```elixir
# Nx.Defn.jit re-traces the function body on every call under compiler: EMLX.
#
#   mix run repro.exs
#
# The compiled program itself is cached; what is not cached is the trace used
# to look it up. Nx passes an opaque `key` to __compile__/4 for exactly this
# purpose, and EMLX ignores it, so `fun.(vars)` runs again on every call and
# the resulting expression is walked to build a structural key.
#
# Cost is proportional to the size of the traced graph.

defmodule Repro do
  import Nx.Defn

  # A chain of `depth` matmuls, each with its own parameter tensor — the shape
  # of a transformer stack, without needing a real model.
  # Cheap ops on purpose: the arithmetic is negligible, so what the timing
  # shows is the cost of walking the graph, not of running it.
  deftransform stack(params, x) do
    params
    |> Tuple.to_list()
    |> Enum.reduce(x, fn w, acc -> acc |> Nx.multiply(w) |> Nx.add(w) end)
  end
end

gpu = fn t -> Nx.backend_transfer(t, {EMLX.Backend, device: :gpu}) end
tpl = fn v -> Nx.Defn.Composite.traverse(v, &Nx.to_template/1) end

bench = fn f ->
  f.(); f.()
  t0 = System.monotonic_time(:microsecond)
  for _ <- 1..50, do: f.()
  (System.monotonic_time(:microsecond) - t0) / 50 / 1000
end

IO.puts("  depth   jit      compile   difference")
IO.puts(String.duplicate("-", 46))

for depth <- [32, 128, 512, 1024] do
  params = for _ <- 1..depth, do: gpu.(Nx.iota({8}, type: :f32))
  params = List.to_tuple(params)
  x = gpu.(Nx.iota({8}, type: :f32))

  jitted = Nx.Defn.jit(&Repro.stack/2, compiler: EMLX)
  compiled = Nx.Defn.compile(&Repro.stack/2, [tpl.(params), tpl.(x)], compiler: EMLX)

  a = bench.(fn -> jitted.(params, x) |> Nx.backend_copy(Nx.BinaryBackend) end)
  b = bench.(fn -> compiled.(params, x) |> Nx.backend_copy(Nx.BinaryBackend) end)

  IO.puts("#{String.pad_leading(to_string(depth), 7)}#{String.pad_leading(Float.round(a, 2) |> to_string(), 8)} ms" <>
    "#{String.pad_leading(Float.round(b, 2) |> to_string(), 9)} ms" <>
    "#{String.pad_leading(Float.round(a - b, 2) |> to_string(), 10)} ms")
end
```

</details>

Codcore:
Closes #136.

`__jit__/5` forwarded straight to `__compile__/4`, which discards the `key` Nx
supplies and calls `native_compile/3`. That runs `fun.(vars)` and walks the
resulting expression on every call — only to arrive at a program already sitting
in the program table.

This caches the closure `__compile__/4` returns, keyed by that `key` plus the
shapes and types of `vars` and the device. On a hit, the trace never runs.

### What is left alone

The program table and its structural key. That key is what lets structurally
identical layers share one compiled program across *different* call sites, and it
still does. The new table sits above it and short-circuits *repeat* calls of the
same call site with the same argument shapes — the shape of every decode loop and
every training step.

Calls carrying `:command_queue` or `:hooks` are not cached: those options
reference caller-owned state, and a process-wide table should not extend its
lifetime. Such calls take the previous path unchanged.

### Measurements

A chain of `depth` cheap ops, so the timing reflects graph traversal rather than
arithmetic (script in #136):

```
  depth   before    after    Nx.Defn.compile
     32   0.59 ms   0.43 ms   0.37 ms
    128   1.01 ms   0.53 ms   0.38 ms
    512   3.12 ms   1.18 ms   0.67 ms
   1024   5.89 ms   1.85 ms   1.30 ms
```

At depth 1024 that is 3.2× faster, and what remains between `jit` and `compile`
is 0.55 ms rather than 4.48 ms.

On a real workload — a 40-layer Qwen3.6-35B-A3B MoE decode step, roughly 4300
nodes — the change removes 8.35 ms per token out of 89.

### Tests

`mix test`: 2668 passed (825 doctests, 1843 tests), 7 excluded. No failures.

### A note on the trade-off

The cache is unbounded, like the program table it sits above. Each entry holds a
closure referencing a compiled program, so a caller that jits the same function
across many distinct argument shapes will accumulate entries. If that is a
concern, the same key would work with a bounded table; I kept it simple to match
the existing behaviour rather than introduce a second eviction policy.

### DIFF
--- a/emlx/lib/emlx.ex
+++ b/emlx/lib/emlx.ex
@@ -1764,9 +1764,75 @@ defmodule EMLX do
   # `dispatch_key/3` still has to fall back to the structural walk on a miss.)
   @dispatch_key_by_id_table :emlx_dispatch_key_by_id
 
+  # Closures returned by `__compile__/4`, keyed by the `key` Nx supplies plus
+  # the shapes/types of `vars` and the options that affect lowering.
+  #
+  # Without this, every `Nx.Defn.jit/2` call re-runs `fun.(vars)` and walks the
+  # resulting expression, only to arrive at a program that is already in
+  # `@program_table`. The program cache stays as it is — its structural key is
+  # what lets structurally identical layers share one program across *different*
+  # call sites. This table is the layer above: it short-circuits *repeat* calls
+  # of the same call site with the same argument shapes, which is the shape of
+  # every decode loop and every training step.
+  @jit_closure_table :emlx_jit_closures
+
   @impl Nx.Defn.Compiler
   def __jit__(key, vars, fun, args_list, opts) do
-    __compile__(key, vars, fun, opts).(args_list)
+    case jit_cache_key(key, vars, opts) do
+      nil ->
+        __compile__(key, vars, fun, opts).(args_list)
+
+      cache_key ->
+        closure =
+          case :ets.lookup(jit_closure_table(), cache_key) do
+            [{^cache_key, cached}] ->
+              cached
+
+            [] ->
+              built = __compile__(key, vars, fun, opts)
+              :ets.insert(jit_closure_table(), {cache_key, built})
+              built
+          end
+
+        closure.(args_list)
+    end
+  end
+
+  # `nil` means "do not cache": an unhashable key, or options carrying a
+  # caller-owned resource whose lifetime we must not extend.
+  defp jit_cache_key(key, vars, opts) do
+    if Keyword.has_key?(opts, :command_queue) or Keyword.has_key?(opts, :hooks) do
+      nil
+    else
+      templates =
+        vars
+        |> Nx.Defn.Composite.flatten_list()
+        |> Enum.map(fn %Nx.Tensor{shape: shape, type: type} -> {shape, type} end)
+
+      {key, templates, Keyword.get(opts, :device, default_device())}
+    end
+  rescue
+    _ -> nil
+  end
+
+  defp jit_closure_table do
+    case :ets.whereis(@jit_closure_table) do
+      :undefined ->
+        try do
+          :ets.new(@jit_closure_table, [
+            :named_table,
+            :public,
+            :set,
+            read_concurrency: true,
+            write_concurrency: true
+          ])
+        rescue
+          ArgumentError -> @jit_closure_table
+        end
+
+      _ ->
+        @jit_closure_table
+    end
   end
 
   @impl Nx.Defn.Compiler
--- a/emlx/lib/emlx.ex
+++ b/emlx/lib/emlx.ex
@@ -1738,112 +1738,57 @@ defmodule EMLX do
 
   @behaviour Nx.Defn.Compiler
 
-  # Known EMLX-specific compiler opts. `:command_queue` is injected by
-  # `__partitions_options__/1` but may also be passed directly by callers
-  # that manage their own queues (equivalent to a manual `with_queue`).
   @valid_compiler_keys [:device, :max_concurrency, :command_queue, :hooks]
 
-  # Process-lifetime dispatch cache backing `dispatch_key/3` +
-  # `get_or_compile_program/6` (see their docs) — a compiled program is keyed
-  # by a *structural* signature of its `Expr` (not object identity), so
-  # it survives across `Nx.Defn.Graph.run/3`'s per-call re-tracing and is
-  # shared across structurally-identical call sites (e.g. every one of
-  # Qwen3's 28 attention layers), not just within one closure's lifetime.
+  # Structural program cache for `get_or_compile_program/6`.
   @native_dispatch_cache_table :emlx_native_dispatch_cache
 
-  # A second, process-lifetime cache in front of `dispatch_key/3`'s own
-  # (expensive — O(nodes), plus per-opaque-scope SHA256 hashing) structural
-  # walk, keyed by `output_expr`'s own node identity rather than its
-  # structural signature. This matters for `run_while_loop/3`'s host-driven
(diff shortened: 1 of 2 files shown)


emlx/lib/emlx.ex
```diff
@@ -1764,9 +1764,75 @@ defmodule EMLX do
...
+          end
+
+        closure.(args_list)
+    end
+  end
+
+  # `nil` means "do not cache": an unhashable key, or options carrying a
+  # caller-owned resource whose lifetime we must not extend.
+  defp jit_cache_key(key, vars, opts) do
+    if Keyword.has_key?(opts, :command_queue) or Keyword.has_key?(opts, :hooks) do
```
polvalente: I think hooks is always a given option. Not sure why this is being filtered on.

emlx/lib/emlx.ex
```diff
@@ -1764,9 +1764,75 @@ defmodule EMLX do
...
+      nil
+    else
+      templates =
+        vars
+        |> Nx.Defn.Composite.flatten_list()
+        |> Enum.map(fn %Nx.Tensor{shape: shape, type: type} -> {shape, type} end)
```
polvalente: You also want to take into account the vectorized_axes and the donation arguments. using Nx.Defn.Composite.traverse with Nx.to_template is likely a better conversion for the arguments. I think the options should also be taken into account because hooks might differ between calls

polvalente (changes requested):

emlx/lib/emlx.ex
polvalente: `:hooks` is always on the jit path, so `has_key?(:hooks)` skipped the cache on every call. The key now stores the hooks map instead of bailing out. Only `:command_queue` still disables the cache, since that is a caller-owned resource.

emlx/lib/emlx.ex
polvalente: The key now builds templates with `Nx.Defn.Composite.traverse/2` and `Nx.to_template/1`, so `vectorized_axes` and names are part of the signature. `:hooks` is in the key as well.

This Nx 0.13 has no donate / `donatable?` field. `to_template/1` will carry that field once it exists.

polvalente (approved):

PR #140, opened 2026-09-02
Add item-count and TTL bounds to the compile-closure cache

### LINKED ISSUE #139: Add bounds to the EMLX compiler cache table
#137 introduced a compilation cache.
We should extend this with a maximum item count, possibly a maximum byte size; we should also add a TTL configuration that uses :infinity by default

Chapaman:
closes #139

#137's `:emlx_compile_closures` table grew without a cap. This adds `compile_cache_max_items` and `compile_cache_ttl` (both `:infinity` by default) so a long-running serving process can opt in without changing today's unbounded behaviour.

### DIFF
--- a/emlx/README.md
+++ b/emlx/README.md
@@ -60,6 +60,19 @@ Nx.Defn.default_options(compiler: EMLX)
 config :nx, :default_defn_options, compiler: EMLX
 ```
 
+EMLX caches jit/compile closures in an ETS table so repeated calls of the same
+function and shapes skip retrace. The table is unbounded by default. For a
+long-running serving process you can cap it:
+
+```elixir
+config :emlx, compile_cache_max_items: 1024          # or :infinity
+config :emlx, compile_cache_ttl: :timer.minutes(30)  # milliseconds, or :infinity
+```
+
+Overflow drops the oldest inserts. TTL is counted from insert time, not last
+access. Evicting a closure does not free the compiled MLX program in the
+native dispatch cache. See the [EMLX moduledoc](https://hexdocs.pm/emlx/EMLX.html#module-compile-once-replay-many-times).
+
 ### Compile-time debug flags
 
 Development-only assertion flags (`:enable_bounds_check`, `:detect_non_finites`,
--- a/emlx/README.md
+++ b/emlx/README.md
@@ -60,18 +60,6 @@ Nx.Defn.default_options(compiler: EMLX)
 config :nx, :default_defn_options, compiler: EMLX
 ```
 
-EMLX caches jit/compile closures in an ETS table so repeated calls of the same
-function and shapes skip retrace. The table is unbounded by default. For a
-long-running serving process you can cap it:
-
-```elixir
-config :emlx, compile_cache_max_items: 1024          # or :infinity
-config :emlx, compile_cache_ttl: :timer.minutes(30)  # milliseconds, or :infinity
-```
-
-Overflow drops the oldest inserts. TTL is counted from insert time, not last
-access. See `EMLX.CompileCache`.
-
 ### Compile-time debug flags
 
 Development-only assertion flags (`:enable_bounds_check`, `:detect_non_finites`,
--- a/emlx/README.md
+++ b/emlx/README.md
@@ -70,8 +70,7 @@ config :emlx, compile_cache_ttl: :timer.minutes(30)  # milliseconds, or :infinit
 ```
 
 Overflow drops the oldest inserts. TTL is counted from insert time, not last
-access. Evicting a closure does not free the compiled MLX program in the
-native dispatch cache. See the [EMLX moduledoc](https://hexdocs.pm/emlx/EMLX.html#module-compile-once-replay-many-times).
+access. See `EMLX.CompileCache`.
 
 ### Compile-time debug flags
 
--- a/emlx/config/config.exs
+++ b/emlx/config/config.exs
@@ -1,5 +1,10 @@
 import Config
 
+# Runtime bounds for the jit/compile eval-closure cache. Uncomment to cap
+# `:emlx_compile_closures`. Both default to :infinity. See config/dev.exs.
+# config :emlx, compile_cache_max_items: 1024
+# config :emlx, compile_cache_ttl: :timer.minutes(30)
+
 if config_env() == :test do
   config :emlx, :add_backend_on_inspect, false
 
--- a/emlx/config/config.exs
+++ b/emlx/config/config.exs
@@ -1,10 +1,5 @@
 import Config
 
-# Runtime bounds for the jit/compile eval-closure cache. Uncomment to cap
-# `:emlx_compile_closures`. Both default to :infinity. See config/dev.exs.
-# config :emlx, compile_cache_max_items: 1024
-# config :emlx, compile_cache_ttl: :timer.minutes(30)
-
 if config_env() == :test do
   config :emlx, :add_backend_on_inspect, false
 
--- a/emlx/config/dev.exs
+++ b/emlx/config/dev.exs
@@ -26,10 +26,3 @@ import Config
 # happen in a working compiler and the checks add wasted work on the hot
 # compile-cache-miss path.
 # config :emlx, compiler_debug: true
-
-# Runtime bounds for `:emlx_compile_closures` (the jit/compile eval-closure
-# cache). Both default to :infinity. TTL is milliseconds from insert time;
-# overflow evicts oldest inserts. Evicting a closure does not free the
-# compiled program in `:emlx_native_dispatch_cache`.
-# config :emlx, compile_cache_max_items: 1024
-# config :emlx, compile_cache_ttl: :timer.minutes(30)
--- a/emlx/config/dev.exs
+++ b/emlx/config/dev.exs
@@ -26,3 +26,10 @@ import Config
 # happen in a working compiler and the checks add wasted work on the hot
 # compile-cache-miss path.
 # config :emlx, compiler_debug: true
+
+# Runtime bounds for `:emlx_compile_closures` (the jit/compile eval-closure
+# cache). Both default to :infinity. TTL is milliseconds from insert time;
+# overflow evicts oldest inserts. Evicting a closure does not free the
+# compiled program in `:emlx_native_dispatch_cache`.
+# config :emlx, compile_cache_max_items: 1024
+# config :emlx, compile_cache_ttl: :timer.minutes(30)
--- a/emlx/lib/emlx.ex
+++ b/emlx/lib/emlx.ex
@@ -146,20 +146,15 @@ defmodule EMLX do
 
   ## Compile once, replay many times
 
-  `Nx.Defn.jit/2` and `Nx.Defn.jit_apply/3` **retrace** the given function
-  from scratch on every call. For most ops that's cheap relative to the
-  actual computation, but for a hot loop (e.g. a decode step, or any call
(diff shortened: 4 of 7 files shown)


emlx/config/config.exs
```diff
@@ -1,5 +1,10 @@
 import Config
 
+# Runtime bounds for the jit/compile eval-closure cache. Uncomment to cap
+# `:emlx_compile_closures`. Both default to :infinity. See config/dev.exs.
+# config :emlx, compile_cache_max_items: 1024
+# config :emlx, compile_cache_ttl: :timer.minutes(30)
+
```
polvalente: Let's remove these, as they're not pulled in by users anyway

emlx/config/dev.exs
```diff
@@ -26,3 +26,10 @@ import Config
 # happen in a working compiler and the checks add wasted work on the hot
 # compile-cache-miss path.
 # config :emlx, compiler_debug: true
+
+# Runtime bounds for `:emlx_compile_closures` (the jit/compile eval-closure
+# cache). Both default to :infinity. TTL is milliseconds from insert time;
+# overflow evicts oldest inserts. Evicting a closure does not free the
+# compiled program in `:emlx_native_dispatch_cache`.
+# config :emlx, compile_cache_max_items: 1024
+# config :emlx, compile_cache_ttl: :timer.minutes(30)
```
polvalente:
```suggestion
```

emlx/config/config.exs
polvalente:
```suggestion
```

emlx/lib/emlx/compile_cache.ex
```diff
@@ -0,0 +1,120 @@
...
+
+      config :emlx, compile_cache_max_items: 1024
+      config :emlx, compile_cache_ttl: :timer.minutes(30)
+
+  Both default to `:infinity`. Overflow evicts the oldest inserts (FIFO).
+  TTL is counted from insert time, not last access. Two concurrent misses
+  on the same key both compile; the later `put/4` wins.
+
+  Evicting a closure does **not** free the compiled MLX program held in
+  `:emlx_native_dispatch_cache`.
```
polvalente: I think it should indeed evict it

emlx/lib/emlx/compile_cache.ex
```diff
@@ -0,0 +1,120 @@
...
+          :ets.delete(table, key)
+          :miss
+        else
+          {:ok, fun}
+        end
+
+      [] ->
+        :miss
+    end
+  end
```
polvalente: I think this module should become a process, such that the TTL expiration logic lives inside a loop of  `Process.send_after(self(), :expire_entries, div(ttl, 2))` + its corresponding `handle_info`

polvalente: Cache reads would only do the lookup with `{:ok, result}` vs `{:error, :cache_miss}`

emlx/lib/emlx/compile_cache.ex
```diff
@@ -0,0 +1,120 @@
...
+      oldest_keys =
+        :ets.foldl(fn {key, _fun, ms}, acc -> [{ms, key} | acc] end, [], table)
+        |> Enum.sort_by(&elem(&1, 0))
+        |> Enum.take(overflow)
+        |> Enum.map(&elem(&1, 1))
+
+      Enum.each(oldest_keys, &:ets.delete(table, &1))
+    else
+      :ok
+    end
```
polvalente: I think `:ets.select_count` could be used to find the amount of items we need to evict. Also if we keep a counter that we insert in the table, we can easily select for the items that should be deleted by matching on `index <= current_counter - max_items` to prune

emlx/lib/emlx.ex
```diff
@@ -146,35 +146,44 @@ defmodule EMLX do
...
+      config :emlx, compile_cache_max_items: 1024
+      config :emlx, compile_cache_ttl: :timer.minutes(30)
+
+  Both accept `:infinity` (the default) or a positive integer. TTL is in
+  milliseconds, counted from insert time — a hit does not refresh it.
+  Overflow evicts the oldest inserts (FIFO). Two concurrent misses on the
+  same key both compile; the later insert wins.
+
+  Evicting a closure does **not** free the compiled MLX program in
+  `:emlx_native_dispatch_cache`. See `EMLX.CompileCache`.
```
polvalente: This docstring seems a bit too verbose, especially since it repeats a lot of what's described in the CompileCache moduledoc

polvalente (commented):
Looking good, but I think we can improve performance!

emlx/lib/emlx/compile_cache.ex
Chapaman: Couldn't drop the native dispatch row here — it's shared across compile-closure keys. Any suggestions?

emlx/lib/emlx/compile_cache.ex
Chapaman: CompileCache is now a GenServer under EMLX.Application. Reads are a plain ETS lookup (`{:ok, fun}` / `{:error, :cache_miss}`). TTL lives in `handle_info(:expire_entries)` on `send_after(div(ttl, 2))`, skipped when ttl is `:infinity` at start (a running sweeper re-reads app env each tick).

emlx/lib/emlx/compile_cache.ex
Chapaman: Switched to an insert index via `:ets.update_counter` and `:ets.select_delete` of `index <= current - max_items`.

emlx/lib/emlx.ex
Chapaman: Dropped the bounds subsection from the EMLX moduledoc; it now points at `EMLX.CompileCache`.

emlx/lib/emlx.ex
```diff
@@ -1859,24 +1855,9 @@ defmodule EMLX do
...
-          :set,
-          read_concurrency: true,
-          write_concurrency: true
-        ])
-
-      _ ->
-        :ok
-    end
+    EMLX.CompileCache.ensure_table()
   end
```
polvalente: We don't need to ensure_table here, we can just trust the default CompileCache init.

polvalente: In fact I think we don't even need this init function

emlx/lib/emlx.ex
```diff
@@ -1915,14 +1896,19 @@ defmodule EMLX do
     cache_key = compile_cache_key(key, vars, hooks, device)
 
     eval_fn =
-      case :ets.lookup(@compile_closure_table, cache_key) do
-        [{^cache_key, cached}] ->
+      case EMLX.CompileCache.fetch(EMLX.CompileCache.table(), cache_key) do
```
polvalente: fetch shouldn't be table-aware, but process aware, defaulting to the global named process

emlx/lib/emlx.ex
```diff
@@ -1915,14 +1896,19 @@ defmodule EMLX do
...
           built = native_compile(vars, fun, device)
-          :ets.insert(@compile_closure_table, {cache_key, built})
-          built
+
+          EMLX.CompileCache.put(
+            EMLX.CompileCache.table(),
+            cache_key,
+            built,
+            EMLX.CompileCache.opts()
+          )
```
polvalente: Same for here, we should just be passing `cache_key` and `built`

emlx/README.md
```diff
@@ -60,6 +60,18 @@ Nx.Defn.default_options(compiler: EMLX)
...
+long-running serving process you can cap it:
+
+```elixir
+config :emlx, compile_cache_max_items: 1024          # or :infinity
+config :emlx, compile_cache_ttl: :timer.minutes(30)  # milliseconds, or :infinity
+```
+
+Overflow drops the oldest inserts. TTL is counted from insert time, not last
+access. See `EMLX.CompileCache`.
+
```
polvalente: Let's remove this from here as it's too much detail for the readme. We can have the moduledoc for EMLX reference EMLX.CompileCache, and the moduledoc from it has the nitty-gritty details

emlx/lib/emlx/compile_cache.ex
```diff
@@ -0,0 +1,157 @@
...
+        n
+
+      other ->
+        raise ArgumentError,
+              "config :emlx, #{env_key(key)} must be :infinity or a positive integer, got: #{inspect(other)}"
+    end
+  end
+
+  defp env_key(:max_items), do: :compile_cache_max_items
+  defp env_key(:ttl), do: :compile_cache_ttl
```
polvalente: how about we use `config :emlx, EMLX.CompileCache, [....]` as the config path?

emlx/lib/emlx/compile_cache.ex
```diff
@@ -0,0 +1,157 @@
...
+    _ = current_ttl(%{ttl: Keyword.get(opts, :ttl)})
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  @doc false
+  def ensure_table(table \\ @table) when is_atom(table) do
+    case :ets.whereis(table) do
+      :undefined ->
+        :ets.new(table, [
+          :named_table,
```
polvalente: The table should not be named, but saved in the process state, and all interactions with it pass through the genserver interface. This serializes access to it, but the process is very lightweight so this isn't really a penalty

polvalente (commented):
Almost there! I think CompileCache should lean more into it's GenServer nature

emlx/lib/emlx/compile_cache.ex
```diff
@@ -21,79 +22,55 @@ defmodule EMLX.CompileCache do
...
-  @doc false
-  def table, do: @table
-
   @doc false
   def start_link(opts \\ []) do
     {name, opts} = Keyword.pop(opts, :name, __MODULE__)
-    _ = current_ttl(%{ttl: Keyword.get(opts, :ttl)})
+    state_opts = %{ttl: Keyword.get(opts, :ttl), max_items: Keyword.get(opts, :max_items)}
+    _ = current_ttl(state_opts)
+    _ = current_max_items(state_opts)
```
polvalente: I think these can be removed right?
Also, let's add Keyword.validate! above?

emlx/lib/emlx/compile_cache.ex
```diff
@@ -21,79 +22,55 @@ defmodule EMLX.CompileCache do
...
-  def put(table \\ @table, key, fun, opts) do
-    max_items = bound!(opts, :max_items)
-    index = :ets.update_counter(table, @counter, 1, {@counter, 0})
-    :ets.insert(table, {key, fun, System.monotonic_time(:millisecond), index})
-    evict_overflow(table, max_items, index)
-    fun
-  end
+  @impl true
+  def init(opts) do
+    table = :ets.new(:emlx_compile_closures, [:set, :private])
```
polvalente: Generally init lives at the top close to start_link

emlx/test/emlx/compile_cache_test.exs
```diff
@@ -4,81 +4,62 @@ defmodule EMLX.CompileCacheTest do
   alias EMLX.CompileCache
 
   setup do
-    table = :ets.new(:compile_cache_test, [:set, :public])
-    %{table: table}
+    name = :"compile_cache_#{System.unique_integer([:positive])}"
+    start_supervised!({CompileCache, name: name, max_items: :infinity, ttl: :infinity}, id: name)
```
polvalente: you could just have it be un-named here and pass the pid around I think

polvalente (approved):

PR #142, opened 2026-09-14
Support Nx 1.0 in the emlx package

tierralibre:
Bumps the `emlx` package from `nx ~> 0.13.1` to `nx ~> 1.0` (lock: nx 1.0.0, complex 1.0.0). No library code changes were needed — the native lowering already speaks `:io_call` — only the test suite needed porting:

- **expr_test**: `Nx.Defn.Kernel.hook/2,3` was renamed to `io_call/2,3` in Nx 1.0, so the test file no longer compiled. Ported the 14 defn call sites plus the 2 cond-branch error-message regexes.
- **linalg doctests**: excluded the new `Nx.LinAlg.matrix_rank/2` doctests. MLX's native SVD returns small-but-nonzero trailing singular values (0.00104 where LAPACK gives exactly 0.0), so the eps-tolerance rank count over-counts on rank-deficient inputs. Same rounding-error class as the existing `svd`/`qr` exclusions.

`emlx_axon` intentionally stays on nx 0.13 in this PR: axon 0.8.1 (latest) requires `nx ~> 0.10` and bumblebee `~> 0.12/0.13`, so that package cannot move until the ecosystem does.

Verification: full `mix test` green — 2687 passed (824 doctests, 1863 tests), 7 excluded (pre-existing excludes only).


emlx/test/emlx/nx_linalg_test.exs
```diff
@@ -16,6 +16,9 @@ defmodule EMLX.Nx.LinalgTest do
   # - lu: MLX raises on singular matrices (doctest uses [[1,2,3],[4,5,6],[7,8,9]])
   # - qr: sign convention differences (-0.0 vs 0.0) and float precision
   # - svd: sign convention differences in U/Vt vectors
+  # - matrix_rank: MLX's native SVD returns small-but-nonzero trailing
+  #   singular values (e.g. 0.00104 where LAPACK gives exactly 0.0), so the
+  #   eps-tolerance rank count over-counts on rank-deficient inputs
```
polvalente:
```suggestion
  # - matrix_rank: MLX's native SVD returns small-but-nonzero trailing
  #   singular values
```

The tests don't rely on LAPACK, but on Nx's custom implementation

polvalente (approved):

polvalente:
Thanks!

